PR #389, opened 2013-06-12
[3.8] Fix #1101 Add sended records throught mail in report

nlegoff:


bburnichon:
to ease follow-up... properly references pr
refs #1101

PR #1101, opened 2014-05-30
[3.8.6][PHRAS-32] Add extended format option for records API responses

nlegoff:
- [x] Add tests


bburnichon:
Ignore previous ref. Linked to another (and unknown) repo

PR #1242, opened 2015-02-17
3.8.7 : Introducing DublinCore metadata template with updated subdef definitions

anthonycabot:
Introducing DublinCore metadata template with updated subdef definitions.


nlegoff:
why did you remove preview_mobie_* sub definitions?

PR #1256, opened 2015-02-23
[3.9] Fix token value sql length

nlegoff:
Some tokens are made of 64 chars, it is truncated un mysql db.


nlegoff:
closed in favor of https://github.com/alchemy-fr/Phraseanet/pull/1257

PR #1284, opened 2015-03-11
Fix search on thesaurus

mdarse:


mdarse:
Change already merged in #1282.

PR #1321, opened 2015-03-25
update

kwemi:


templates/web/prod/index.html.twig
```diff
@@ -528,7 +525,214 @@
 
             <div id="idFrameT" class="PNB ui-corner-top">
                 <div class="tools PNB10 btn-toolbar">
-                    {% include "prod/toolbar.html.twig" with {acl: app['acl'].get(app['authentication'].getUser()) } %}
+
```
bburnichon: Why has this change been reverted? We need this for the plugin architecture.

PR #1352, opened 2015-04-16
api story creation

jygaulier:
PHRAS-29
- fix : title is written into the first "thumbtitle" field
- enhancement : metadata can be set
- enhancement : check the ContentType is application/json


bburnichon:
:+1:

PR #1374, opened 2015-05-07
report des routes api/story de 3.8

jygaulier:
#PHRAS-485
#time 1.5d


lib/Alchemy/Phrasea/ControllerProvider/Api/V1.php
```diff
@@ -1363,7 +1473,7 @@ protected function add_record_to_story(Application $app, \record_adapter $story,
 
         try {
             $record = new \record_adapter($app, $databox_id, $record_id);
-        } catch (\Exception_Record_AdapterNotFound $e) {
+        } catch (Exception_Record_AdapterNotFound $e) {
```
bburnichon: Leading backslash is mandatory in this case. Old non namespaced classes are not imported in current namespace.

bburnichon:
sniff... I already refactored this controller in #1360 but it is not merged yet.

PR #1382, opened 2015-05-20
Refactor merged field structure

mdarse:
Much more extensive testing is needed.

ping @bburnichon can you review this?

### DIFF
--- a/lib/Alchemy/Phrasea/SearchEngine/Elastic/Indexer/RecordIndexer.php
+++ b/lib/Alchemy/Phrasea/SearchEngine/Elastic/Indexer/RecordIndexer.php
@@ -28,6 +28,8 @@ use Alchemy\Phrasea\SearchEngine\Elastic\Indexer\RecordQueuer;
 use Alchemy\Phrasea\SearchEngine\Elastic\Mapping;
 use Alchemy\Phrasea\SearchEngine\Elastic\RecordHelper;
 use Alchemy\Phrasea\SearchEngine\Elastic\StringUtils;
+use Alchemy\Phrasea\SearchEngine\Elastic\Structure\Field;
+use Alchemy\Phrasea\SearchEngine\Elastic\Structure\Structure;
 use Alchemy\Phrasea\SearchEngine\Elastic\Thesaurus;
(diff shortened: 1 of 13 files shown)


lib/Alchemy/Phrasea/SearchEngine/Elastic/Indexer/RecordIndexer.php
```diff
@@ -55,8 +57,9 @@ class RecordIndexer
 
     private $logger;
 
-    public function __construct(RecordHelper $helper, Thesaurus $thesaurus, \appbox $appbox, array $locales, LoggerInterface $logger)
+    public function __construct(Structure $structure, RecordHelper $helper, Thesaurus $thesaurus, \appbox $appbox, array $locales, LoggerInterface $logger)
     {
+        $this->structure = $structure;
```
bburnichon: missing structure property declaration.

lib/Alchemy/Phrasea/SearchEngine/Elastic/Structure/Structure.php
```diff
@@ -0,0 +1,99 @@
...
+
+    public function typeOf($name)
+    {
+        return isset($this->fields[$name]) ?
+                     $this->fields[$name]->getType() : null;
+    }
+
+    public function isPrivate($name)
+    {
+        return isset($this->private[$name]) ? true :
```
bburnichon: method name induced me that return type is `bool`. Actually it is `bool|null`. I usually check returned values with triple equal operator. It can become hard to know that this method returned null instead of false.

If the field does not exists and could not be found, I would prefer method to throw exception instead. WDYT?

bburnichon:
It definitely needs more testing. Only one small test! But otherwise, it improves code.

Current test suite is broken. Should be fixed before going further.

lib/Alchemy/Phrasea/SearchEngine/Elastic/Structure/Structure.php
mdarse: It was using exceptions at first, but wasn’t sure about consistency with `get()` and `typeOf()`. But yeah, it feels strange to have two « falsy » values meaning different things.

mdarse:
thanks @bburnichon

PR #1383, opened 2015-05-21
Add media accesor

bburnichon:
PHRAS-519


lib/Alchemy/Phrasea/Controller/Api/V1Controller.php
```diff
@@ -921,9 +933,37 @@ private function listEmbeddableMedia(\record_adapter $record, \media_subdef $med
...
+        ];
+        if ($url_ttl >= 0) {
+            $payload['exp'] = $payload['iat'] + $url_ttl;
+        }
+
+        /** @var SecretProvider $provider */
+        $provider = $this->app['provider.secrets'];
+        $secret = $provider->getSecretForUser($issuer);
+
+        return sprintf('media_accessor', [
```
bburnichon: oups... should not be an `sprintf`

Issue #1334, opened 2015-04-02
Split Controller providers and controllers.


bburnichon:
Very big PR to Split controller providers into proper `ControllerProvider` class and `Controller` class

- [x] Admin/Collection
- [x] Admin/ConnectedUsers
- [x] Admin/Dashboard
- [x] Admin/Databox
- [x] Admin/Databoxes
- [x] Admin/Fields
- [x] Admin/Publications
- [x] Admin/Root
- [x] Admin/SearchEngine
- [x] Admin/Setup
- [x] Admin/Subdefs
- [x] Admin/TaskManager
- [x] Admin/Users
- [x] Api/Oauth2
- [x] Api/V1
- [x] Client/Root
- [x] Datafiles
- [x] Lightbox
- [x] Minifier
- [x] Permalink
- [x] Prod/BasketController
- [x] Prod/Bridge
- [x] Prod/DoDownload
- [x] Prod/Download
- [x] Prod/Edit
- [x] Prod/Export
- [x] Prod/Feed
- [x] Prod/Language
- [x] Prod/Lazaret
- [x] Prod/MoveCollection
- [x] Prod/Order
- [x] Prod/Printer
- [x] Prod/Property
- [x] Prod/Push
- [x] Prod/Query
- [x] Prod/Records
- [x] Prod/Root
- [x] Prod/Share
- [x] Prod/Story
- [x] Prod/TOU
- [x] Prod/Tools
- [x] Prod/Tooltip
- [x] Prod/Upload
- [x] Prod/UsrLists
- [x] Prod/WorkZone
- [x] Report/Activity
- [x] Report/Informations
- [x] Report/Root
- [x] Root/Account
- [x] Root/Developers
- [x] Root/Login
- [x] Root/RSSFeeds
- [x] Root/Root
- [x] Root/Session
- [x] Setup
- [x] Thesaurus/Thesaurus
- [x] Thesaurus/Xmlhttp
- [x] User/Notifications
- [x] User/Preferences

bburnichon:
Linked to PHRAS-529

PR #1393, opened 2015-06-09
Phras 368 th reindex evt

jygaulier:
PHRAS-368
thesaurus modification events
cs


bburnichon:
OK, I should be careful while refactoring controllers for #1334. :-)

PR #1395, opened 2015-06-09
PHRAS-442 #time 5h

jygaulier:
facets returned in API
"suggestions" returned in API (as flat facets)

### DIFF
--- a/lib/Alchemy/Phrasea/Controller/Api/V1Controller.php
+++ b/lib/Alchemy/Phrasea/Controller/Api/V1Controller.php
@@ -1099,19 +1099,19 @@ class V1Controller extends Controller
         $this->getSearchEngine()->clearCache();
 
         $ret = [
-            'offset_start'      => $offsetStart,
-            'per_page'          => $perPage,
+            'offset_start' => $offsetStart,
+            'per_page' => $perPage,
             'available_results' => $search_result->getAvailable(),
-            'total_results'     => $search_result->getTotal(),
-            'error'             => (string)$search_result->getError(),
-            'warning'           => (string)$search_result->getWarning(),
-            'query_time'        => $search_result->getDuration(),
-            'search_indexes'    => $search_result->getIndexes(),
-            'suggestions'       => array_map(
+            'total_results' => $search_result->getTotal(),
+            'error' => (string)$search_result->getError(),
+            'warning' => (string)$search_result->getWarning(),
+            'query_time' => $search_result->getDuration(),
+            'search_indexes' => $search_result->getIndexes(),
+            'suggestions' => array_map(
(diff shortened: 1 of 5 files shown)


lib/Alchemy/Phrasea/Controller/Api/V1Controller.php
```diff
@@ -1081,6 +1081,7 @@ private function prepareSearchRequest(Request $request)
                 function (SearchEngineSuggestion $suggestion) {
                     return $suggestion->toArray();
                 }, $search_result->getSuggestions()->toArray()),
+            'facets'            => $search_result->getFacets(),
```
bburnichon: Can you please remove superfluous spaces? Thanks

lib/Alchemy/Phrasea/SearchEngine/Elastic/Search/FacetsResponse.php
```diff
@@ -31,6 +31,11 @@ public function __construct(Escaper $escaper, array $response)
         }
     }
 
+    public function getFacets()
```
bburnichon: Please add phpdoc block to ease IDE in guessing types.

bburnichon:
Seems you broke 2 tests.

lib/Alchemy/Phrasea/SearchEngine/Elastic/Search/FacetsResponse.php
```diff
@@ -31,6 +31,16 @@ public function __construct(Escaper $escaper, array $response)
...
     }
 
+    /**
+     * Term aggregates
+     *
+     * @return array
+     */
+    public function getFacets()
+    {
+        return $this->facets;
```
mdarse: You should not do that. This is private structure, if you expose it we will not be able to change it anymore.

lib/Alchemy/Phrasea/SearchEngine/Elastic/ElasticSearchEngine.php
```diff
@@ -292,8 +293,16 @@ public function query($string, $offset, $perPage, SearchEngineOptions $options =
...
         }
 
+        /** @var FacetsResponse $facets */
         $facets = $this->facetsResponseFactory->__invoke($res);
 
+        // for es, suggestions are a flat view of facets (api backward compatibility)
+        $suggestions->clear();
+        foreach($facets->getFacets() as $facet) {
+            foreach($facet['values'] as $value) {
+                $suggestions->add(new SearchEngineSuggestion($value['query'], $value['value'], $value['count']));
```
mdarse: This code should not be here, at least it should be extracted in another method or class, `ElasticSearchEngine` is too heavy already. Maybe this belong in some `FacetResponse::getSuggestions()` method.

PR #1399, opened 2015-06-15
Fix IN search regression

mdarse:
Introduced in 40e2df6c3f8653ff482f3cd8165cfbea9461f7e5.


mdarse:
It look like this patch also fix regression on equality operator :)

bburnichon:
thanks.

PR #1405, opened 2015-06-19
PHRAS-593 fixes video, pdf and audio preview in prod interface

lostdalek:


bburnichon:
I think it is risky to rely on a random value for an id.
Is this macro used in a lot of places? The instability should be extracted to the caller responsibility. WDYT?

bburnichon:
OK, replace this would be too much work. Merging as add JIRA todo

PR #1407, opened 2015-06-19
Fix type filter regression

mdarse:


bburnichon:
I could have bet it was a simple bug.

PR #1408, opened 2015-06-22
Empty indexer queues after flushing to index

mdarse:
I use here SplObjectStorage::removeAllExcept() with an empty one since I didn't
found a simpler one step solution.


lib/Alchemy/Phrasea/SearchEngine/Elastic/Indexer.php
```diff
@@ -191,6 +191,10 @@ public function flushQueue()
             $this->recordIndexer->delete($bulk, $this->deleteQueue);
             $bulk->flush();
         });
+
+        static $empty = new SplObjectStorage();
```
bburnichon: `$this->indexQueue = new \SplObjectStorage();`

bburnichon:
The simplest way to remove all elements from a `SplObjectStorage` is to actually create a new one.

PR #1419, opened 2015-07-01
PHRAS-482 #time 4h

jygaulier:
- added "recordid:xxx"
- added "base:yyy" (makes "Base" facets ok)
to query parser


lib/Alchemy/Phrasea/SearchEngine/Elastic/Search/QueryVisitor.php
```diff
@@ -262,4 +268,24 @@ private function visitCollectionNode(Element $element)
 
         return new AST\CollectionExpression($collectionName);
     }
+
+    private function visitrecordidNode(Element $element)
```
bburnichon: Should be `visitRecordidNode`.

PR #1423, opened 2015-07-03
Restrict search on visible private fields

mdarse:
We only allow querying on private fields where the user has edit right.


mdarse:
@bburnichon Can you review and merge this please?

PR #1425, opened 2015-07-07
Add API get collection route

aztech-dev:


lib/Alchemy/Phrasea/Controller/Api/V1Controller.php
```diff
@@ -466,6 +466,13 @@ private function listDatabox(\databox $databox)
         ];
     }
 
+    public function getDataboxCollectionAction(Request $request, $base_id)
+    {
+        return Result::create($request, [
+            $this->listCollection($this->app['phraseanet.appbox']->get_collection($base_id))
```
bburnichon: Please use `$this->app->getApplicationBox()` instead

lib/Alchemy/Phrasea/Application/Api.php
```diff
@@ -92,7 +92,9 @@
         if ($request->getRequestFormat(Result::FORMAT_JSON) === Result::FORMAT_JSONP && !$response->isOk() && !$response->isServerError()) {
             $response->setStatusCode(200);
         }
-        // set response content type
+        // set response content typeReturns available collections on a specified databox.
```
bburnichon: Typo?

lib/classes/appbox.php
```diff
@@ -367,7 +367,7 @@ public function get_databoxes()
 
         return $this->databoxes;
     }
-
+
```
bburnichon: strange, why are there spaces here?

PR #1428, opened 2015-07-09
Clean boxes

aztech-dev:
Extract repository/factory from collection


bburnichon:
Ouch, that is a big PR... Will need some time to review it. :-)

PR #1463, opened 2015-07-29
#PHRAS-622 #time 3h

jygaulier:
add 'lenient' to prevent es errors when parsing invalid dates
cs


mdarse:
This do not fix the root cause of the problem, this is an issue with the date fields being included in search even if the query contains an invalid date.
Issue is here https://github.com/alchemy-fr/Phraseanet/blob/3569d4746746af3bca8b5099d0c04f63994cd0b2/lib/Alchemy/Phrasea/SearchEngine/Elastic/Structure/Field.php#L117

cc @bburnichon & @jygaulier

PR #1480, opened 2015-08-26
Wrap text and concept queries with a single filtered query

mdarse:


mdarse:
@bburnichon this is ready to be reviewed & merged :)

PR #1497, opened 2015-09-07
Fix cross field queries (see PHRAS-691)

mdarse:


mdarse:
@bburnichon I fixed tests!

PR #1503, opened 2015-09-16
PHRAS-716 #time 3d

jygaulier:
fix : admin/base progress bar (indexation)
fix : reindex button
fix : indexation does not skip records between fetches


bburnichon:
An `array_flip` causes tests to fail. I don't know if it is also case in a full environment.

Also, there is a bunch of code duplicated at least 3 times. It would be nice to wrap this in a method.

PR #1550, opened 2015-10-29
PHRAS-776 - fix acl issues

lostdalek:


lib/Alchemy/Phrasea/Controller/Prod/Tools.php
```diff
@@ -50,12 +50,19 @@ public function connect(Application $app)
...
+
                 foreach ($subdefs as $subdef) {
-                    $permalink = $subdef->get_permalink();
-                    $recordAccessibleSubdefs[] = array(
-                      'name' => $subdef->get_name(),
-                      'state' => $permalink->get_is_activated()
-                    );
+                    $subdefName = $subdef->get_name();
+
+                    if ($acl->has_access_to_subdef($record, $subdefName) && $acl->is_admin()) {
```
bburnichon: isn't it an OR instead of AND?

lostdalek:
it's an AND, some subdefs shouldn't be displayed because they are not desactivable even for an admin user.

PR #1590, opened 2015-12-07
PHRAS-759 - embed bundle

lostdalek:


lib/Alchemy/Phrasea/Application.php
```diff
@@ -13,6 +13,7 @@
 
 use Alchemy\Cors\Options\DefaultProvider;
 use Alchemy\CorsProvider\CorsServiceProvider;
+use Alchemy\EmbedProvider\EmbedServiceProvider;
```
bburnichon: Useless line. Should be removed

PR #1608, opened 2015-12-22
Phras 882 base indexable

jygaulier:


lib/Alchemy/Phrasea/TaskManager/Job/IndexerJob.php
```diff
@@ -45,6 +48,17 @@ public function getEditor()
...
+        $indexer = $app['elasticsearch.indexer'];
+
+        foreach (
+            array_filter(
+                $app->get_databoxes(),
+                function($databox) use($app) {
+                    return $app->is_databox_indexable($databox);
+                }
+            ) as $databox) {
+            $indexer->indexScheduledRecords($databox);
```
bburnichon: Should be
```php
$databoxes = array_filter($app->get_databoxes(), function (\databox $databox) use ($app) {
    return $app->is_databox_indexable($databox);
});
$indexer->indexScheduledRecords($databoxes);
```

You can also add a method to `DataboxRepository::findIndexable(): \databox[]`

PR #1609, opened 2015-12-23
#PHRAS-874

jygaulier:
time  3h
fix : sql error when no status-bits are defined in structure
fix : last high-order bit(s) where not reset in table "status"
bad side effect : 28 (32-4) entries in table "status" for each record.


bburnichon:
fix via #1610

PR #1628, opened 2016-01-20
Profiler improvements

aztech-dev:
Adds a new WebProfilerServiceProvider that provides the following services:

- Ajax profiling tab (not available in Silex provider for v1.0)
- Custom Doctrine profiler init (other service provider does play along nice)
- Cache profiling tab


lib/Alchemy/Phrasea/Core/Provider/WebProfilerServiceProvider.php
```diff
@@ -0,0 +1,133 @@
...
+            $cache->setNamespace($namespace);
+
+            return $cache;
+        }));
+
+        $app['data_collector.cache_subscriber'] = $app->share(function ($app) {
+            return new CacheStatisticsSubscriber($app['cache']);
+        });
+
+        $app->share($app->extend('data_collectors', function ($collectors) use ($app) {
```
bburnichon: `$app['data_collectors'] = $app->share($app->extends('data_collectors', ...));`

lib/Alchemy/Phrasea/Core/Provider/WebProfilerServiceProvider.php
```diff
@@ -0,0 +1,133 @@
...
+            $collectors['db'] = $app->share(function ($app) {
```
bburnichon: That's `$collectors` you are extending, so `$collectors['db'] = $collectors->share(function () use ($app) {`

aztech-dev: $collectors is an array, not a Pimple container...

lib/Alchemy/Phrasea/Core/Provider/WebProfilerServiceProvider.php
aztech-dev: $collectors is an array, not a Pimple container...

templates-profiler/cache.html.twig
```diff
@@ -0,0 +1,109 @@
...
+                <th>Key</th>
+                <th>Hit / Miss</th>
+            </tr>
+        </thead>
+        <tbody>
+        {% for call in collector.calls %}
+            <tr>
+                <td>{{ loop.index }}</td>
+                <td>
+                    {% if call['type'] != 'fetch' %}<strong>{% endif %}
```
bburnichon:
```twig
{% if call['type'] != 'fetch' %}
  <strong>{{ call['type'] }}</strong>
{% else %}
  {{ call['type'] }}
{% endif %}
```

lib/Alchemy/Phrasea/Core/Provider/WebProfilerServiceProvider.php
bburnichon: oups, false guess...

You should still assign `$app->share()` to `$app['data_collectors']`

lib/Alchemy/Phrasea/Core/Provider/WebProfilerServiceProvider.php
bburnichon: Missing assignment

PR #1631, opened 2016-01-20
Cleanup ORM bootstrap

aztech-dev:
PHRAS-915

- Removes dependency on external ORM service provider (simpler code)
- Removes all unnecessary public ORM services except where needed for BC


lib/Alchemy/Phrasea/Controller/SetupController.php
```diff
@@ -140,16 +140,17 @@ public function doInstall(Request $request)
             ]);
         }
 
-        $this->app['dbs.options'] = array_merge(
+        /*$this->app['dbs.options'] = array_merge(
```
bburnichon: Why not remove this block?

lib/Alchemy/Phrasea/Core/Provider/ORMServiceProvider.php
```diff
@@ -12,570 +12,207 @@
...
-                    throw new \UnexpectedValueException(sprintf('Unknown driver "%s"', $params['driver']));
-            }
-        });
 
-        // Hash a DSN string
-        $app['hash.dsn'] = $app->protect(function($dsn) {
-            return md5($dsn);
-        });
+    private static $customTypes = [
+        'blob' => '\Alchemy\Phrasea\Model\Types\Blob',
```
bburnichon: use `::class` constructs? Even `Types\Blob::class` to only import `Alchemy\Phrasea\Model\Types` namespace.

lib/Alchemy/Phrasea/Core/Provider/ORMServiceProvider.php
```diff
@@ -12,570 +12,207 @@
...
         });
 
-        //Return DoctrineServiceProvider information for a database from its parameters
-        $app['db.options.from_info'] = $app->protect(function(array $info) use ($app) {
-            $info = $app['db.info']($info);
-
-            $key = $app['hash.dsn']($app['db.dsn']($info));
-
-            return array($key => $info);
+        $app['connection.pool.manager'] = $app->share(function ($app) {
```
bburnichon: Is this service used at many places? Seems pretty useless as is.

aztech-dev: Yes without it you get "too many connections" errors, because the connection factory is used in a lot of places with the assumption that it always returns the same connection for a given set of parameters.

lib/Alchemy/Phrasea/Core/Provider/ORMServiceProvider.php
```diff
@@ -12,570 +12,207 @@
...
-                            throw new InvalidArgumentException(sprintf('Missing "%s" argument for database connection using driver %s', $param, $info['driver']));
-                        }
+    private function validateConnectionSettings(array $parameters)
+    {
+        if (!isset($parameters['driver'])) {
+            $parameters['driver'] = 'pdo_mysql';
+        }
+
+        if (!isset($parameters['charset'])) {
+            $parameters['charset'] = 'utf8';
```
bburnichon: SQLite does not know about this charset encoding.
The Recommended for MySQL is utf8mb4.

lib/Alchemy/Phrasea/Core/Provider/ORMServiceProvider.php
bburnichon: There are 8 references in whole project to `$app['connection.pool.manager']`. Can we rename these calls as `$app['dbal.connection_pool']` instead?

lib/Alchemy/Phrasea/Setup/Installer.php
```diff
@@ -144,7 +144,7 @@ private function rollbackInstall(Connection $abConn, Connection $dbConn = null)
     private function createAB(Connection $abConn)
     {
         // set default orm to the application box
-        $this->app['orm.ems.default'] = $this->app['hash.dsn']($this->app['db.dsn']($abConn->getParams()));
+        //$this->app['orm.ems.default'] = $this->app['hash.dsn']($this->app['db.dsn']($abConn->getParams()));
```
bburnichon: delete with useless comment

lib/Alchemy/Phrasea/Core/Provider/ORMServiceProvider.php
aztech-dev: It will be fixed in an upcoming P/R

aztech-dev:
Closing this P/R in favor of smaller, more targeted P/R's

PR #1646, opened 2016-01-27
Avoid many calls to cache layer

bburnichon:
Getting collection calls access_restriction which is not properly cached.

Remove use of cache in AccessRestriction and use instance memory cache instead.
Beware static keyword declares variable static for the class, not the instance

### DIFF
--- a/lib/classes/collection.php
+++ b/lib/classes/collection.php
@@ -188,20 +188,57 @@ class collection implements ThumbnailedElement, cache_cacheableInterface
             ));
         }
 
-        $repository = self::getRepository($app, $reference->getDataboxId());
-        $collection = $repository->find($reference->getCollectionId());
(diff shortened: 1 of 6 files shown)


lib/classes/collection.php
```diff
@@ -188,52 +188,73 @@ public static function getByBaseId(Application $app, $base_id)
...
-        }
+    /**
+     * @param  Application $app
+     * @param  databox $databox
+     * @param  int $collectionId
+     * @return collection
+     */
+    public static function getByCollectionId(Application $app, databox $databox, $collectionId)
+    {
+        assert(is_int($collectionId));
```
aztech-dev: Why not use the assertions provided by beberlei since we use already use them in the code base ?

bburnichon: This assert call was already there. There are several uses of `assert()` calls which should be converted.

PR #1647, opened 2016-01-27
#PHRAS-920 time 1h

jygaulier:
fix : "BaseFromColl" (dbox coll_id -> abox base_id) check for sbas_id


aztech-dev:
:+1: Sorry for the bug

PR #1653, opened 2016-01-28
Configure event processor factory via DI

aztech-dev:
Allows registering more event processor factories


jygaulier:
service is not registered in app

PR #1679, opened 2016-02-11
PHRAS-601 - fix behavior in delete records modal

lostdalek:


bburnichon:
merged via f3824d5600becca6b0614ac521b327337018b142

PR #1681, opened 2016-02-11
Remove "install" tests from Travis suite

aztech-dev:
They are already run on CircleCI, so running them on Travis is redundant. This should also help reducing build times, and freeing up the queue faster for other builds.


bburnichon:
:+1:

PR #1700, opened 2016-02-16
Unsaved user preferences

bburnichon:
PHRAS-980


bburnichon:
close in favor of #1701

PR #1704, opened 2016-02-17
PHRAS-974 - PHRAS-974 - fix publication module

lostdalek:


lostdalek:
contain PHRAS-972 too

PR #1730, opened 2016-02-25
API create order

bburnichon:
PHRAS-1001


lib/Alchemy/Phrasea/Collection/Reference/DbalCollectionReferenceRepository.php
```diff
@@ -99,6 +112,29 @@ public function findByCollectionId($databoxId, $collectionId)
         return null;
     }
 
+    public function findHavingOrderMaster(array $subset = null)
```
aztech-dev: It would be nicer to call this parameter collectionSubset for better readibility

PR #1933, opened 2016-08-16
PHRAS-1178_edit-story-and-records

jygaulier:
### Fixes
  cache is now invalidated after editing records
  recordeditsuscriber can't run into recursion anymore


aztech-dev:
On en fait quoi ?

PR #1941, opened 2016-09-21
PHRAS-842_search-logged-as-es

jygaulier:
### Fixes
  - PHRAS-842: query is logged in a "user" form, before parsing


jygaulier:
@aztech-dev : pr in master (sorry) because the 4.0 fails on my vm (?)

PR #1981, opened 2016-10-13
PHRAS-1236 Enable native session handlers

aztech-dev:
### Fixes
- PHRAS-1236: It was not possible to use session handlers other than "files"


aztech-dev:
Waiting for validation before merge

PR #2015, opened 2016-11-15
Change lib captcha for password brute force attack

xrousset78800:
### Adds
  - PHRAS-1268 - Change lib captcha and add trial number in admin config


aztech-dev:
Not good, all dependencies have been updated in this P/R with incompatible package versions given our current minimum supported version

PR #2026, opened 2016-12-05
Change lib captcha on login and register form

xrousset78800:
### Adds
  - PHRAS-1268 Change lib captcha on login and register form
  - Composer.lock is missing (facebook-sdk issue)


templates/web/login/register-classic.html.twig
```diff
@@ -82,7 +82,7 @@
 
                     {{ form_rest(form) }}
 
-                    {% if recaptcha_display %}
+                    {% if recaptcha_enabled %}
```
aztech-dev: Same comment as https://github.com/alchemy-fr/Phraseanet/pull/2026/files#diff-c41b766e77e45067195d812e3f3bc24bR110

templates/web/login/layout/base-layout.html.twig
```diff
@@ -39,6 +39,7 @@
                 <script type="text/javascript" src="/assets/vendors/json2/json2{% if not app.debug %}.min{% endif %}.js"></script>
                 <![endif]-->
                 <script type="text/javascript" src="/assets/vendors/modernizr/modernizr{% if not app.debug %}.min{% endif %}.js"></script>
+                <script src="https://www.google.com/recaptcha/api.js" async defer></script>
```
aztech-dev: The library should installed via Bower or NPM and deployed in public assets instead of adding it directly from a CDN

lib/Alchemy/Phrasea/Application.php
```diff
@@ -418,7 +416,7 @@ public function getUnlockAccountData()
      */
     public function requireCaptcha()
     {
-        if ($this['conf']->get(['registry', 'webservices', 'captcha-enabled'])) {
+        if ($this['conf']->get(['registry', 'webservices', 'captchas-enabled'])) {
```
aztech-dev: Why are you renaming an existing configuration key ? It will break configurations for existing installs

lib/Alchemy/Phrasea/Core/Provider/AuthenticationManagerServiceProvider.php
```diff
@@ -139,14 +145,16 @@ public function register(Application $app)
         });
 
         $app['auth.native'] = $app->share(function (Application $app) {
-            $authConf = $app['conf']->get('authentication');
+            $authConf = $app['conf']->get('registry');
+
+            if ($authConf['webservices']['captchas-enabled']) {
```
aztech-dev: Why are you changing the configuration key ?

nmaillat:
closing, PR out of date

PR #2072, opened 2017-03-01
PHRAS-1343_DEFINE-FILENAME-AT-DONWNLOAD

jygaulier:
### Adds
  - PHRAS-1343: allow to specify a filename on permalink usage


jygaulier:
add "filename=foo.ext" as argument to permalink : ...?token=1234&download=1&filename=foo.jpg

PR #2078, opened 2017-04-04
PHRAS-1388_Invalid-geonameId

moctardiouf:
### Fixes
  - PHRAS-1388: message "invalid geoname id" is not displayed anymore for empty fields


jygaulier:
this is a new relayed to an already merged fix (bburnichon). different reason but same symptom.

PR #2102, opened 2017-05-18
PHRAS-1349_bin/console fields:rename

yannick-esokia:
### Changed
  lib/classes/module/console/fieldsRename.php
  
### Fixes
  PHRAS-1349_bin/console fields:rename
  Fixed bug


nmaillat:
Bad workflow

PR #2117, opened 2017-05-30
PHRAS-1378_Bin_console_add_max_result_window_parameter

yannick-esokia:
### Changed
  - lib/Alchemy/Phrasea/Controller/Admin/SearchEngineController.php
  
### Fixes
  - PHRAS-1378_Bin_console_add_max_result_window_parameter
  - Bug Result window is too large


jygaulier:
The new parameter should be passed using the "options" (add methods "set/getMaxResultWindow()") and used by the "createIndex()" method.
Every call to es should be done via the api, no direct call, no hardcoded values.

yannick-esokia:
The correction is in https://github.com/alchemy-fr/Phraseanet/pull/2121

PR #2121, opened 2017-05-31
PHRAS-1378_Bin_console_add_max_result_window_parameter

yannick-esokia:
### Changed
  - lib/Alchemy/Phrasea/Command/SearchEngine/IndexCreateCommand.php
  
### Fixes
  PHRAS-1378_Bin_console_add_max_result_window_parameter
  Bug Result window is too large


jygaulier:
The new method "dumpResultIndexElasticsearchAction()" is not related to this ticket but to 1404.
btw, if possible (check php-es-api ?) the "dumpResultIndexElasticsearchAction()" should not do direct call to es.

PR #2183, opened 2017-07-06
PHRAS-1455 - Create basket from items in order

falylala:
### Adds
  - PHRAS-1455: Create basket from items in order
  - Add new route to create basket from items in order


nmaillat:
duplicate PR

PR #2188, opened 2017-07-10
Phras 1449 add options populate command

falylala:
### Adds
  - PHRAS-1449: Bin/console searchengine:index 
  -  Add limit and order option for command populate


nmaillat:
Close in favor of 
https://github.com/alchemy-fr/Phraseanet/pull/2182

PR #2225, opened 2017-08-09
PHRAS-1513 standard menus

falylala:
### Adds
  - PHRAS-1513: Style menu
  - Add new class for menu

### Removes
 - PHRAS-1513: Style menu
 -  Remove divider between  and space items


nmaillat:
won't do this feature in this version

PR #2330, opened 2017-10-24
PHRAS-1360-Truncate table memcached

falylala:
### Adds
  - PHRAS-1360: Add patch 
  - Truncate table memcached


nmaillat:
PHRAS-1360 won't do like this

PR #2443, opened 2018-01-18
PHRAS-1825 Update signature year to 2018

xrousset78800:
### Changed
  - PHRAS-1825 Update signature year to 2018.


nmaillat:
close because this is to late

PR #2463, opened 2018-02-08
PHRAS-1860: fix push_list field type to text

moctardiouf:
### Fixes
  - PHRAS-1860: fix push_list field type
  - fix push_list field type from varchar(255) to text


nmaillat:
PHRAS-1860 #comment close in favor of pr #3226

PR #2489, opened 2018-02-23
PHRAS-569 Add "reasons" in api/v1/quarantine

aynsix:
### Adds
  - PHRAS-569: Add record_id in /api/v1/quarantine


jygaulier:
RecordID is not good.
An uploaded FILE has NO record id.
An uploaded file can conflict for MANY reasons=Checkers, ex. : 
- bad colorspace (NOT related to any record in db).
- same uuid as record(s) (related to 1 or MANY records in db)
The idea is to get the same informations as shown  in prod/upload (reason, message) for every failing checker.
I propose an array "Conflicts" with the checker as key, giving the reason/message and an array of record_ids if relevant).

PR #2524, opened 2018-03-16
PHRAS-1911-add route return list documentary fields

KallooaSan:
### Adds
  - PHRAS-1911: add route return list documentary fields
  - refactor : added static function getTechnicalFieldsList()
  - added tests

### DIFF
--- a/lib/classes/media/subdef.php
+++ b/lib/classes/media/subdef.php
@@ -77,6 +77,9 @@ class media_subdef extends media_abstract implements cache_cacheableInterface
     /** @var integer */
     private $size = 0;
 
+    /** @var array */
+    private static $technicalFieldsList = [];
+
     /*
      * Players types constants
      */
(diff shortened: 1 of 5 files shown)


lib/classes/media/subdef.php
```diff
@@ -825,4 +803,42 @@ private function save()
...
     }
+
+    /**
+     * Return list of technical data and there attributes
+     *
+     * @return array
+     */
+    public static function getTechnicalFieldsList(){
+
+        $datas = [
```
jygaulier: datas should better be declared (in the method or in the class) as "private static $technicalFieldsList..." so it is initialized only once in case of multiple calls to getTech...()
When it's not possible to initialize the static to a runtime-only value, initialize it with null, false, empty or whatever, and set the value only once with a if(isnull($x)).
Also : because 'name' is always set and unique, it can be used as the array key, and test like array_key_exists('name'... can be avoided.
btw the whole list could be in a silex/symfony container but that's another question, it's ok for now.

PR #2539, opened 2018-03-29
PHRAS-774 fix tokens create account

KallooaSan:
### Fixes
  - PHRAS-774 : fix tokens create account
  - Add delay of one month before purge
  - Renew token on expiration

### DIFF
--- a/lib/Alchemy/Phrasea/Model/Repositories/TokenRepository.php
+++ b/lib/Alchemy/Phrasea/Model/Repositories/TokenRepository.php
@@ -68,8 +68,25 @@ class TokenRepository extends EntityRepository
                 WHERE t.expiration < :date';
 
         $query = $this->_em->createQuery($dql);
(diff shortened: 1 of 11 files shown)


lib/Alchemy/Phrasea/Model/Repositories/TokenRepository.php
```diff
@@ -68,8 +68,25 @@ public function findExpiredTokens()
                 WHERE t.expiration < :date';
 
         $query = $this->_em->createQuery($dql);
-        $query->setParameters([':date' => new \DateTime()]);
+        $query->setParameters([':date' => new \DateTime('-1 month')]);
```
jygaulier: no. The expiration date must be absolute, there is no reason to add a 1 month delta in a "findExpiredTokens()" method.
btw in the next method "findExpiredToken()" there is no delta.
todo : check that the expiration is SET as now()+1 month on creation.

PR #2550, opened 2018-04-18
PHRAS-1809_Geoinfos-in-index

jygaulier:
### Adds
  - PHRAS-1809: Add a "location" field in records (type="geo_point"), populated with same values as "Longitude" and "Latitude" technical fields.


lib/Alchemy/Phrasea/SearchEngine/Elastic/Indexer/RecordIndex.php
```diff
@@ -73,6 +73,8 @@ public function getMapping()
         $mapping->addIntegerField('height')->disableIndexing();
         $mapping->addIntegerField('size')->disableIndexing();
 
+        $mapping->addGeoPointField('location')->disableIndexing();
```
jygaulier: do NOT disableIndexing, else kibana :( 
ALREADY fixed in master afert this pr was merged.

PR #2564, opened 2018-05-02
fix circle build (gulp)

jygaulier:


jygaulier:
close, same as PHRAS-2010 (better name)

PR #2662, opened 2018-07-05
PHRAS-2086_AdobeCC-AppToken_4.1

jygaulier:
### Adds
  - PHRAS-2086: add an "adobe_cc-plugin" api application
require system:upgrade to go to 4.1.0-alpha.10
must be enabled via "adobe_cc-enabled" in conf

### DIFF
--- a/lib/Alchemy/Phrasea/Model/Manipulator/ApiApplicationManipulator.php
+++ b/lib/Alchemy/Phrasea/Model/Manipulator/ApiApplicationManipulator.php
@@ -51,6 +51,13 @@ class ApiApplicationManipulator implements ManipulatorInterface
         return $application;
     }
 
+    public function findByClientID($client_id)
+    {
+        return $this->repository->findOneBy([
+            'clientId' => $client_id
+        ]);
+    }
+
     public function delete(ApiApplication $application)
     {
         $this->om->remove($application);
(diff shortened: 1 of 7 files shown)


lib/classes/patch/410alpha10a.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public function apply(base $databox, Application $app)
+    {
+        // create an api application for adobeCC
+
+        /** @var ApiApplicationManipulator $manipulator */
+        $manipulator = $app['manipulator.api-application'];
+
+        if($manipulator->findByClientID(\API_OAuth2_Application_adobeCCPlugin::CLIENT_ID) === null) {
+
```
jygaulier: better to use
$app['repo.api-applications']->findByClientId(), so no need to add the method in manipulator

lib/Alchemy/Phrasea/Model/Manipulator/ApiApplicationManipulator.php
```diff
@@ -51,6 +51,13 @@ public function create($name, $type, $description, $applicationWebsite, User $cr
         return $application;
     }
 
+    public function findByClientID($client_id)
+    {
+        return $this->repository->findOneBy([
+            'clientId' => $client_id
+        ]);
+    }
```
jygaulier: tocheck : useless ? method exists in app['repo.api-applications']

PR #2734, opened 2018-09-11
PHRAS-2196 API / Stories record pagination on search answer and Stories fetch info

aynsix:
## Changelog

The "/api/v3/search" endpoint, in story mode, limits the number of records returned for each stories.
By default, only 10 records are returned for each story, but this is can be set with parameter "story_max_records".
This endpoint has a same behavior as /api/v1/search when search in record mode

The "/api/v3/stories" endpoint is now paginated with two parameters

 - per_page : the number of records per page, default value is 10 
 - page: the desired page, default is page one


lib/Alchemy/Phrasea/Controller/Api/V1Controller.php
```diff
@@ -1178,7 +1179,8 @@ public function searchAction(Request $request)
         $searchView = $this->buildSearchView(
             $result,
             $includeResolver->resolve($fractal),
-            $this->resolveSubdefUrlTTL($request)
+            $this->resolveSubdefUrlTTL($request),
+            $request
```
jygaulier: The whole request should not be passed to a function that does not need it (for building a response). "buildSearchView()" should not know it's called from an (http)action. Ideally, request should only appear on route handlers "xxxAction()". I know that "request" is sent everywhere (easy), it's not a big deal on private methods that build final response, but here, it's nicer to compute "max_items", "page" into the handler "searchAction()", and then pass them to "buildSearchView()"
Just aesthetic...

lib/Alchemy/Phrasea/Controller/Api/V1Controller.php
```diff
@@ -1253,6 +1256,12 @@ private function buildSearchView(SearchEngineResult $result, array $includes, $u
         $resultView = new SearchResultView($result);
 
         if ($stories->count() > 0) {
+            $storie_max_items = null;
+
+            if($request->getAcceptableContentTypes() == V1::$extendedContentTypes['json']){
+                $storie_max_items = (int)$request->get('storie_max_items')?:10;
```
nmaillat: story_max_items better than storie_max_item

lib/Alchemy/Phrasea/Databox/Record/LegacyRecordRepository.php
```diff
@@ -129,25 +129,72 @@ public function findChildren(array $storyIds, $user = null)
...
-        ;
+        if($max_items){
+            array_unshift($selects, 'sr.rid_parent as story_id');
+
+            $subBuilder = $connection->createQueryBuilder();
+
+            $subBuilder
+                ->select('s.*,
+                    IF(@old_rid_parent != s.rid_parent, @cpt := 1, @cpt := @cpt+1) AS CPT')
+                ->addSelect("IF(@old_rid_parent != s.rid_parent, IF(@old_rid_parent:=s.rid_parent,'NEW PARENT',0), '----------') AS Y")
```
jygaulier: this "Y" was for debug, it can be removed

lib/Alchemy/Phrasea/Databox/Record/LegacyRecordRepository.php
```diff
@@ -129,25 +129,72 @@ public function findChildren(array $storyIds, $user = null)
...
+            $builder->select($selects)
+                ->from(sprintf('( %s )', $subBuilder->getSQL()), 'sr')
+                ->innerJoin('sr', 'record', 'r', 'r.record_id = sr.rid_child')
+                ->where('sr.CPT BETWEEN :offset AND :maxresult')
+                ->andWhere('r.parent_record_id = 0')
+                ->setParameter('offset', $offset)
+                ->setParameter('maxresult', ($offset + $max_items -1))
+                ->orderBy('story_id, sr.CPT')
+            ;
+
```
jygaulier: the query can be built with one builder since this sql runs ok :
```
SET @cpt=1;
SET @old_rid_parent=-1;
SELECT
    r.coll_id AS collection_id,
    r.record_id,
    r.credate AS created,
    r.uuid,
    r.moddate AS updated,
    r.parent_record_id AS isStory,
    r.type,
    r.originalname AS originalName,
    r.sha256,
    r.mime,
    LPAD(BIN(r.status), 32, '0') as status
    , rid_parent as story_id
FROM
(
    SELECT
        r.*, g.rid_parent,
        IF(@old_rid_parent != rid_parent, @cpt := 1, @cpt := @cpt+1) AS CPT
    FROM
        (regroup AS g INNER JOIN record AS r ON(r.record_id=g.rid_child))
    ORDER BY
        rid_parent,
        ord
)
AS
    r
WHERE
    CPT BETWEEN 1 AND 5
```

PR #2806, opened 2018-11-08
PHRAS-2255 Add totals of records in a story, for route "search" (story mode), "stories" and "stories/embed"

aynsix:
### Adds
  - PHRAS-2255: Add totals of records in a story, for route "search" (story mode), "stories" and "stories/embed"
  - add total_story_items in results of api route /search and /stories


nmaillat:
Close in favor of pr #3268

PR #2808, opened 2018-11-12
PHRAS-2289 API search route add a "page" parameter but keep offset_start=

aynsix:
### Adds
  - PHRAS-2289: API search route add a "page" parameter but keep offset_start= in "api/v3/search/" endpoint


nmaillat:
close in favor of pr #3611 and #3613

PR #2819, opened 2018-11-22
PHRAS-2276 port to 4.1 template escape

aynsix:
### Fixes
  - PHRAS-2276: port to 4.1 template escape


lib/Alchemy/Phrasea/Controller/Prod/PushController.php
```diff
@@ -597,7 +597,7 @@ public function editListAction(Request $request, $list_id)
 
     private function formatUser(User $user)
     {
-        $subtitle = array_filter([$user->getJob(), $user->getCompany()]);
+        $subtitle = array_filter([htmlspecialchars($user->getJob()), htmlspecialchars($user->getCompany())]);
```
jygaulier: No. A method in the controller does not have to know what usage is made of the result (html), this is the job of the final consumer to format for his own usage.
Here "formatUser" is called by a few public methods that return json, json should not contain html.
Html escaping should be done at render-time, usually in twig.

lib/Alchemy/Phrasea/Controller/Thesaurus/ThesaurusController.php
```diff
@@ -2128,7 +2128,7 @@ private function getXMLTerm($bid, $id, $typ, $piv, $sortsy, $sel, $nots, $acf)
                                 $sy = $sy_list->appendChild($ret->createElement("sy"));
 
                                 $sy->setAttribute("id", $n->getAttribute("id"));
-                                $sy->setAttribute("v", $t = $n->getAttribute("v"));
+                                $sy->setAttribute("v", htmlspecialchars($t = $n->getAttribute("v")));
```
jygaulier: Same as above : we should not pre-escape for xml (setAttribute does it). We should only escape when we generate html.

lib/classes/record/adapter.php
```diff
@@ -941,7 +941,7 @@ public function getTitle($locale = null, Array $options = [])
             $this->set_data_to_cache(self::CACHE_TITLE, $title);
         }
 
-        return $title;
+        return htmlspecialchars($title);
```
jygaulier: NO ! NEVER do encoding at low level. Do it ONLY on the usage, for eg. in the twig.

PR #2836, opened 2018-11-29
PHRAS-2316 add log on setmetadata route api

aynsix:
### Adds
  - PHRAS-2316:  add log on setmetadata route api


jygaulier:
will be done on api v3

PR #2918, opened 2019-02-08
PHRAS-1972: Account Page, Allow a Phraseanet User to delete his account and associated datas

aynsix:
### Adds
  - PHRAS-1972:  Account Page, Allow a Phraseanet User to delete his account and associated datas
  - first step of allow a user to delete account


lib/Alchemy/Phrasea/Controller/Root/AccountController.php
```diff
@@ -299,11 +305,99 @@ public function displayAccount()
...
         return $this->render('account/account.html.twig', [
-            'user'          => $user,
-            'evt_mngr'      => $manager,
-            'notifications' => $manager->list_notifications_available($user),
+            'user'                => $user,
+            'evt_mngr'            => $manager,
+            'notifications'       => $manager->list_notifications_available($user),
+            'baskets'             => $baskets,
+            'api_accounts'        => $apiAccounts,
+            'isOwnerPublisher'    => count($ownerPublisher) ? true : false,
```
jygaulier: better : 'isOwnerPublisher'  => count($ownerPublisher) > 0;

lib/Alchemy/Phrasea/Controller/Root/AccountController.php
```diff
@@ -299,11 +305,99 @@ public function displayAccount()
...
+            $response->headers->clearCookie('last_act');
+
+            return $response;
+        }
+
+    }
+
+    public function confirmDeleteAccount(Request $request)
+    {
+        if (null !== $tokenValue = $request->query->get('token')) {
```
jygaulier: AGAIN : NO YODA. This is NOT an option.
if( ($tokenValue = $request->query->get('token')) !== null)

lib/Alchemy/Phrasea/Controller/Root/AccountController.php
```diff
@@ -299,13 +306,102 @@ public function displayAccount()
         $manager = $this->getEventManager();
         $user = $this->getAuthenticatedUser();
 
+        $repo_baskets = $this->getBasketRepository();
+        $baskets = $repo_baskets->findActiveValidationAndBasketByUser($user);
+
+        $apiAccounts = $this->getApiAccountRepository()->findByUser($user);
+
+        $ownerPublisher = $this->getFeedPublisherRepository()->findBy(['user' => $user, 'owner' => true]);
```
jygaulier: semantic : rename as $ownedFeeds to match the type of objects found in repo.

lib/Alchemy/Phrasea/Controller/Root/AccountController.php
```diff
@@ -299,13 +306,102 @@ public function displayAccount()
...
+
+        $ownerFeedbackSend = $this->getValidationSessionRepository()->findby(['initiator' => $user, ]);
```
jygaulier: rename as $initiatedValidations

lib/Alchemy/Phrasea/Controller/Root/AccountController.php
```diff
@@ -299,13 +306,102 @@ public function displayAccount()
...
-            'user'          => $user,
-            'evt_mngr'      => $manager,
-            'notifications' => $manager->list_notifications_available($user),
+            'user'                => $user,
+            'evt_mngr'            => $manager,
+            'notifications'       => $manager->list_notifications_available($user),
+            'baskets'             => $baskets,
+            'api_accounts'        => $apiAccounts,
+            'isOwnerPublisher'    => count($ownerPublisher) > 0 ? true : false,
+            'isOwnerFeedbackSend' => count($ownerFeedbackSend) > 0 ? true : false,
```
jygaulier: For previous args ($user, $manager, ...), the whole object is passed to twig which is responsible to use the object.
It would be better to do the same : pass $ownedFeeds, $initiatedValidations, and let twig do the if(count(...)), because if gives you the chance to display things like "you own X feeds", not only a bool.

btw : comparison ops (<,>,==,<=,...) are booleans, so there is no need to add ternary.

lib/Alchemy/Phrasea/Controller/Root/AccountController.php
aynsix: done

lib/Alchemy/Phrasea/Controller/Root/AccountController.php
aynsix: done

lib/Alchemy/Phrasea/Controller/Root/AccountController.php
aynsix: ok

templates/web/account/account.html.twig
```diff
@@ -239,7 +239,7 @@
 
                         <div style="max-width: 400px;margin-left: 1%;width: 47%;display: inline-block;margin-right: 2%;float: right;">
 
-                            {% if not user.isadmin() and not user.hasLdapCreated() and not isOwnerPublisher and not isOwnerFeedbackSend %}
+                            {% if not user.isadmin() and not user.hasLdapCreated() and not owned_feeds and not initiated_validations %}
```
jygaulier: where is the count() ?

PR #2929, opened 2019-02-26
PRAS-2457_Prod_Search_Form__The_color_stays_bleu_whether_a_filter_is_…

HRavalomanana:
### Fixes
* PRAS-2457: Prod Search Form : The color stays bleu whether a filter is active or not
   it is the PR from the clean branch instead of https://github.com/alchemy-fr/Phraseanet/pull/2928

### DIFF
--- a/resources/www/prod/js/jquery.main-prod.js
+++ b/resources/www/prod/js/jquery.main-prod.js
@@ -155,15 +155,6 @@ function checkFilters(save) {
     var statusFilters = $('#ADVSRCH_SB_ZONE .status-section-title .danger_indicator', container);
     var dateFilterSelect = $('#ADVSRCH_DATE_ZONE select', container);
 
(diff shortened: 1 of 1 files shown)


resources/www/prod/js/jquery.main-prod.js
```diff
@@ -151,9 +151,19 @@ function checkFilters(save) {
...
 
+    // Manage Fields status
+    statusField.removeClass('danger');
+    $.each(fieldsSelectFake, function(index,el){
+        if( $(el).val() !== '' ) {
+            danger = true;
+            statusField.addClass('danger');
+        }
+    });
+
```
jygaulier: there is already a fieldSelectFake loop later in the code, better to set "danger" there.

nmaillat:
Fix PHRAS-2457

PR #2933, opened 2019-03-05
WIP: PHRAS-2290_Prod__push_or_feedback__add-user__generates_a_javascript_e…

HRavalomanana:
### Fixes
  - PHRAS-2290: Prod / push or feedback / add-user / generates a javascript error when geonames is null
- The corrected and clean PR is : #2938

### DIFF
--- a/templates/web/prod/User/Add.html.twig
+++ b/templates/web/prod/User/Add.html.twig
@@ -47,7 +47,7 @@
                     <label for="">{{ 'City' | trans }}</label>
                 </td>
                 <td>
-                    <input name="city" type="text" value="" class="geoname_field"/>
+                    <input name="city" type="text" value="" class="{% if app['geonames.server-uri'] %}geoname_field{% endif %}"/>
                 </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


templates/web/prod/User/Add.html.twig
```diff
@@ -108,11 +108,13 @@
...
-        "server": "{{ app['geonames.server-uri'] }}",
-        "limit": 40
-    });
+    geolink = '{{ app['geonames.server-uri'] }}';
+    if(geolink.length > 0) {
+        $addUserForm.find('.geoname_field').geocompleter({
+            "server": "{{ app['geonames.server-uri'] }}",
+            "limit": 40
+        });
+    }
```
jygaulier: since app['geo...'] is a constant from the twig point of view, it's better to do a "twig:if" and not generate the useless js code at all.

HRavalomanana: It is corrected

jygaulier: the js code is still generated with "server": "".
The "if" could wrap the whole js code, not the class.

HRavalomanana: This part of code is from the 'geonames-server-jquery-plugin/jquery.geonames.js' whitch is a library called via gulp so i can't edit it. Unless if we will call this file separately ?

HRavalomanana:
The corrected and clean PR is : https://github.com/alchemy-fr/Phraseanet/pull/2938

PR #2942, opened 2019-03-08
PHRAS-2184 : List manager - advance mode - email info is missing in filter box

aynsix:
### Adds
  - PHRAS-2184: List manager - advance mode - email info is missing in filter box
  - Add email domain filter in the first box


aynsix:
moved to https://github.com/alchemy-fr/Phraseanet/pull/3448

PR #2970, opened 2019-04-11
PHRAS-2517 : Worker service in 4.0

aynsix:
## Changelog

- PHRAS-2517: update phraseanet to allow using plugin service worker


nmaillat:
PHRAS-2517 #comment need review form JYG
#time 4h testing port code from 4.1 to 4.0

lib/Alchemy/Phrasea/Controller/Prod/ExportController.php
```diff
@@ -191,40 +192,68 @@ public function exportMail(Request $request)
...
-            \set_export::build_zip(
-                $this->app,
-                $token,
-                $list,
-                $this->app['tmp.download.path'].'/'. $token->getValue() . '.zip'
-            );
 
-            $remaingEmails = $destMails;
+            //  if the service worker plugin is installed and enabled , use it
+            if ($this->app['plugins.manager']->isEnabled('phraseanet-plugin-services')) {
```
jygaulier: no. the core code must not contain "if(x) then event() else hardcode()".
The app should not test if a plugin is installed (app must not know plugins).
The job is event-driven or hardcoded, not both.
As event-driven, the controller simply dispatch a core (internal) event - that's all -, and a core event-listener does the hardcoded job.
When the "plugin-service" is installed, it overrides the core listener and does his own way.

lib/Alchemy/Phrasea/Controller/Prod/ToolsController.php
```diff
@@ -165,6 +166,8 @@ public function imageAction(Request $request)
 
             if (!$substituted || $force) {
                 $record->rebuild_subdefs();
+
+                $this->dispatch(RecordEvents::SUB_DEFINITION_REBUILD, new SubDefinitionRebuildEvent($record));
```
jygaulier: We should not do rebuild() AND send event (even if rebuild() is only raising a status-bit for a task). We should send event TO rebuild. 
2 options to send events :
- high level, into controllers. Each time a code did call "rebuild_subdefs()" (7 calls in ph), it now sends an event. The "rebuild_subdefs()" is removed.
- low level, into "rebuild_subdefs()". Send the event from here.

Each case the hard code (building subdefs) is moved to a listener. The default listener will simply raise the sb, but a plugin might override the listener and run a worker.

I prefer option 1 since it allows to remove code from the core.

nmaillat:
won't do

PR #2996, opened 2019-04-25
PHRAS-2507 Enhance AddPlugin add DownloadPlugin command

MilosEsokia:
### Changed
  - bin/setup plugins:add command detects if source is remote and switches on new plugins:download command

### Adds
  -  bin/setup plugins:download command downloads remote .zip file or clone git repo from remote source to local

### DIFF
--- a/lib/Alchemy/Phrasea/Command/Plugin/AbstractPluginCommand.php
+++ b/lib/Alchemy/Phrasea/Command/Plugin/AbstractPluginCommand.php
@@ -51,4 +51,43 @@ abstract class AbstractPluginCommand extends Command
         $this->container['plugins.autoloader-generator']->write($manifests);
         $output->writeln(" <comment>OK</comment>");
     }
+
+    protected function validateSource($source)
+    {
+        $valid_scheme = false;
+        $valid_extension = false;
+
+        $allowed_scheme = array('http','https','ftp');
+        $allowed_extension = array('zip','git');
+
+        $scheme =  parse_url($source, PHP_URL_SCHEME);
+        if (in_array($scheme, $allowed_scheme)){
+            $valid_scheme = true;
+        }
+
+        $path = parse_url($source, PHP_URL_PATH);
+        if (strpos($path, '.') !== false) {
+            $path_parts = explode('.', $path);
+            $extension = $path_parts[1];
+            if (in_array($extension, $allowed_extension)){
+                $valid_extension = true;
+            }
+        }
+
+        if ($valid_scheme && $valid_extension){
+            return $extension;
+        } else {
+            return false;
+        }
+    }
+
+    public static function delDirTree($dir) {
+        $files = array_diff(scandir($dir), array('.','..'));
+        foreach ($files as $file) {
+            (is_dir("$dir/$file")) ? self::delDirTree("$dir/$file") : unlink("$dir/$file");
+        }
+        return rmdir($dir);
+    }
+
+
 }
--- a/lib/Alchemy/Phrasea/Command/Plugin/AbstractPluginCommand.php
+++ b/lib/Alchemy/Phrasea/Command/Plugin/AbstractPluginCommand.php
@@ -57,7 +57,7 @@ abstract class AbstractPluginCommand extends Command
         $valid_scheme = false;
         $valid_extension = false;
 
-        $allowed_scheme = array('http','https','ftp');
+        $allowed_scheme = array('https','ssh');
         $allowed_extension = array('zip','git');
 
         $scheme =  parse_url($source, PHP_URL_SCHEME);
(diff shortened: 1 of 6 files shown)


lib/Alchemy/Phrasea/Command/Plugin/AbstractPluginCommand.php
```diff
@@ -51,4 +51,43 @@ protected function updateConfigFiles(InputInterface $input, OutputInterface $out
...
+        }
+
+        if ($valid_scheme && $valid_extension){
+            return $extension;
+        } else {
+            return false;
+        }
+    }
+
+    public static function delDirTree($dir) {
```
4rthem: why public?

lib/Alchemy/Phrasea/Command/Plugin/DownloadPlugin.php
```diff
@@ -0,0 +1,129 @@
...
+                    $output->writeln("Downloading <info>$source</info>...");
+                    $repo = GitRepository::cloneRepository($source, $local_download_path);
+                    $output->writeln("Plugin downloaded to <info>$local_download_path</info>");
+                    break;
+
+            }
+
+        } else {
+
+            $output->writeln("The source <info>$source</info> is not valid remote(.zip or .git)");
```
4rthem: Actually it could be valid but not supported:

```suggestion
            $output->writeln(sprintf('The source <info>%s</info> is supported. Only .zip and .git are supported.', $source));
```

lib/Alchemy/Phrasea/Command/Plugin/DownloadPlugin.php
```diff
@@ -0,0 +1,129 @@
...
+        $this
+            ->setDescription('Downloads a plugin to Phraseanet')
+            ->addArgument('source', InputArgument::REQUIRED, 'The source is a remote url (.zip or .git)')
+            ->addArgument('destination', InputArgument::OPTIONAL, 'Download destination');
+    }
+
+
+    protected function doExecutePluginAction(InputInterface $input, OutputInterface $output)
+    {
+
```
4rthem: Please respect coding standards (camelCase, remove trailing spaces, etc...)

lib/Alchemy/Phrasea/Command/Plugin/DownloadPlugin.php
```diff
@@ -0,0 +1,129 @@
...
+        $source = $input->getArgument('source');
+        $destination = $input->getArgument('destination');
+
+        $destination_subdir = '/plugin-'.time();
```
4rthem: you can have conflict with `time()` (if many plugins are installed automatically at the same time for instance).
You can `md5()` the source.

lib/Alchemy/Phrasea/Command/Plugin/DownloadPlugin.php
```diff
@@ -0,0 +1,129 @@
...
+
+        if ($destination){
+
+            $destination = trim($destination);
+            $destination = ltrim($destination, '/');
```
4rthem: Why do you remove the first `/`?

lib/Alchemy/Phrasea/Command/Plugin/DownloadPlugin.php
```diff
@@ -0,0 +1,129 @@
...
+            $destination = rtrim($destination, '/');
+
+            $local_download_path = $destination;
+
+        } else {
+
+            $local_download_path = 'tmp/plugin-download' . $destination_subdir;
```
4rthem: On unix, it is `/tmp`...

lib/Alchemy/Phrasea/Command/Plugin/AbstractPluginCommand.php
MilosEsokia: Copy - Paste from other project

lib/Alchemy/Phrasea/Command/Plugin/DownloadPlugin.php
MilosEsokia: I thought downloaded plugin should be stored in Phraseanet code tree

lib/Alchemy/Phrasea/Command/Plugin/DownloadPlugin.php
MilosEsokia: Right, same as above

lib/Alchemy/Phrasea/Command/Plugin/DownloadPlugin.php
4rthem: I still don't understand why. You cannot transform an absolute path into a relative one like this.

PR #3005, opened 2019-05-07
PHRAS-2529 Enable console debug depending on environment

MilosEsokia:
### Changed
  - CLI.php changed in order ot create logger only when in dev environment or environment set in configuration - node example: 
'console_logger_enabled_environments' => [ 'test','prod' ]
  
### Fixes
  - PHRAS-2529: Remove SQL request screen dump from bin/console bin/setup and bin/developer


lib/Alchemy/Phrasea/CLI.php
```diff
@@ -73,9 +73,26 @@ public function __construct($name, $version = null, $environment = self::ENV_PRO
...
-        $this['logger'] = $this->extend('logger', function () {
-            return new Console\Logger\ConsoleLogger(new Console\Output\ConsoleOutput(Console\Output\ConsoleOutput::VERBOSITY_DEBUG));
-        });
+        $logger = false;
+
+        if ($this['configuration.store']->isSetup()){
+
+            $config = $this['configuration.store']->getConfig();
+
+            if ((isset($config['console_logger_enabled_environments']) && in_array($environment, $config['console_logger_enabled_environments'])) || $environment == self::ENV_DEV){
```
jygaulier: testing env_dev here is useless

MilosEsokia: As I understood the request, console logger should be activated either when in dev environment or in environment set in some configuration file. Should just configuration be checked? So if user set just 'prod', logger will not be activated when in dev?

MilosEsokia: I thought that logger should always be activated if in dev

jygaulier: yes, in dev logger is alway on. Since it's tested on line 87 anyway, we don't need a "or" on line 82.

MilosEsokia: Sorry, since I update it few times I forgot, you are completely right - will remove || $environment == self::ENV_DEV clause

PR #3014, opened 2019-05-14
PHRAS-2399 fix connexion timeout leading to record insert error

moctardiouf:
### Fixes
  - PHRAS-2399: Archive Task Generate error when connection to db is lost
  - fix connexion timeout leading to record insert error


nmaillat:
Close in favor #3022

PR #3039, opened 2019-05-17
Revert "PHRAS-2546 : add monolog handler"

nmaillat:
Reverts alchemy-fr/Phraseanet#3007


nmaillat:
we don't need to revert anymore

PR #3051, opened 2019-05-24
PHRAS-2572 : add route POST api/v1/upload/enqueue

aynsix:
### Adds
  - PHRAS-2572: add route POST api/v1/upload/enqueue on Phraseanet


aynsix:
move the route api/v1/upload/enqueue  in the plugin Phraseanet-plugin-service

PR #3098, opened 2019-07-12
PHRAS-2665 Fix video tool ratio

HRavalomanana:
### Fixes
  - PHRAS-2665: Prod-video tools - set image - Use preview ratio instead document ratio

### DIFF
--- a/templates/web/prod/actions/Tools/videoEditor.html.twig
+++ b/templates/web/prod/actions/Tools/videoEditor.html.twig
@@ -49,6 +49,7 @@
                                 {#Set the preview Ratio #}
                                 {% set width = technical_info[dataW].value %}
                                 {% set height = technical_info[dataH].value %}
+                                {% set prevRatio = '' %}
 
                                 {% if width and height %}
                                     {% set rawRatio =  (width / height)|number_format(3, '.') %}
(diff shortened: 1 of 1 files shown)


templates/web/prod/actions/Tools/videoEditor.html.twig
```diff
@@ -56,12 +59,30 @@
...
+                                {% for subdef in previewHtml5 %}
+
+                                    {% set width = subdef.get_width()  %}
+                                    {% set height = subdef.get_height() %}
+
+                                    {% if width and height %}
+                                        {% set rawRatio =  (width / height)|number_format(3, '.') %}
+                                        {% set rawRatioLength =  rawRatio|length %}
+                                        {% if subdef.get_name()=='preview' %}
+                                            {% set prevRatio = rawRatio|slice(0,rawRatioLength-1) %}
```
jygaulier: 2 decimals will always be enough, no need for 3 then slice to 2...
btw : removing last digit is wrong since 1.248 should become 1.25

templates/web/prod/actions/Tools/videoEditor.html.twig
```diff
@@ -56,12 +59,30 @@
...
+                                        {% endif %}
+                                    {% else %}
+                                        {% set prevRatio = '' %}
```
jygaulier: adding data-ratio="" may have side effects (?), it's better not to add the attribute at all.

HRavalomanana: data-ratio is needed on javascript part to generate the canvas size

PR #3100, opened 2019-07-15
PHRAS-2651 event to message 4.1

jygaulier:
!!! WIP DO NOT MERGE !!!

### Adds
  - PHRAS-2651: allow serialisation of events to push on queue and re-post later


nmaillat:
PHRAS-2651 won't do like this

PR #3120, opened 2019-07-26
PHRAS-2691 Prod -delete from trash , send deletion by bulk of 3 records

HRavalomanana:
### Changed
  - Prod -delete from trash , send deletion by bulk of 3 records


nmaillat:
PHRAS-2691 #comment close in favor of PR #3134

PR #3127, opened 2019-08-01
PHRAS-1069 : Dates seems not extracted from iptc

aynsix:
### Fixes
  - PHRAS-1069: Dates seems not extracted from iptc
  - format date from metadata to Y/m/d H:i:s

### DIFF
--- a/lib/Alchemy/Phrasea/Metadata/PhraseanetMetadataSetter.php
+++ b/lib/Alchemy/Phrasea/Metadata/PhraseanetMetadataSetter.php
@@ -67,7 +67,10 @@ class PhraseanetMetadataSetter
                 }
 
                 if ($field->get_type() == 'date') {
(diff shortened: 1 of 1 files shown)


lib/Alchemy/Phrasea/Metadata/PhraseanetMetadataSetter.php
```diff
@@ -66,8 +66,18 @@ public function replaceMetadata($metadataCollection, \record_adapter $record)
                     continue;
                 }
 
-                $data['value'] = $value;
+                if ($field->get_type() == 'date') {
+                    $dateTime = date_create($value);
```
jygaulier: utiliser Datetime avec try/catch

aynsix: ok

PR #3159, opened 2019-09-04
Phras 2729 datetime warnings 4.0

jygaulier:
### Fixes
  - PHRAS-2729: a date value like "Film de Jan De BONT (1997)" was exploded as ["","1997",""], now non-numeric parts are excluded


nmaillat:
PHRAS-2729 #comment Merge of  datetime warnings 4.0

PR #3173, opened 2019-09-20
PHRAS-2741 : worker service part 1

aynsix:
## Changelog

- PHRAS-2741 : split the worker service


lib/Alchemy/Phrasea/Border/Manager.php
```diff
@@ -14,6 +14,8 @@
 use Alchemy\Phrasea\Application;
 use Alchemy\Phrasea\Border\Checker\CheckerInterface;
 use Alchemy\Phrasea\Border\Attribute\AttributeInterface;
+use Alchemy\Phrasea\Core\Event\Record\RecordEvents;
+use Alchemy\Phrasea\Core\Event\Record\SubdefinitionBuildEvent;
```
jygaulier: change (everywhere) "build" to "create", since other events are "...creation", "...created"

PR #3195, opened 2019-10-16
PHRAS-2389 Update fixtures.sql to be compliant with strict mode

xrousset78800:
### Fixes
  - PHRAS-2389 | Update fixtures.sql to propose a default value to each field. As Strict mode is requiring for.


nmaillat:
close in favor off pr #3342

PR #3249, opened 2019-12-02
PHRAS-2603 : Migration patch for configuration.yml and tag a new Phraseanet Version

aynsix:
### Fixes
  - PHRAS-2603: Migration patch for configuration.yml and tag a new Phraseanet Version
  #comment 
  - Fix embed-bundle key
  - replace 'name' to 'map-provider' in geoloc section
  - remove registry classic section
  - remove bridge section if exist
  - insert RGPD bloc if not exist


lib/Alchemy/Phrasea/Core/Version.php
```diff
@@ -16,7 +16,7 @@ class Version
     /**
      * @var string
      */
-    private $number = '4.1.0-alpha.19a';
+    private $number = '4.1.0-alpha.20a';
```
nmaillat: Need to bump version to 4.1.0-alpha.23a

lib/classes/patch/410alpha20a.php
```diff
@@ -0,0 +1,111 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Alchemy\Phrasea\Application;
+
+class patch_410alpha20a implements patchInterface
+{
+    /** @var string */
+    private $release = '4.1.0-alpha.20a';
```
nmaillat: bump version to 4.1.0-alpha.23a and rename file

PR #3259, opened 2019-12-10
PHRAS-2852 : fix error on circleci WIP 4.1

aynsix:
### Fixes
  - PHRAS-2852:  fix error on circleci

### DIFF
--- a/tests/Alchemy/Tests/Phrasea/Controller/MinifierTest.php
+++ b/tests/Alchemy/Tests/Phrasea/Controller/MinifierTest.php
@@ -49,13 +49,21 @@ class MinifierTest extends \PhraseanetTestCase
 
     public function testFileMinificationWithUnknownGroupShouldReturnA500()
     {
-        self::$DI['client']->request('GET', '/include/minify/?g=prout');
-        $this->assertEquals(500, self::$DI['client']->getResponse()->getStatusCode());
+        // there is an exception thrown
+        try {
+            $this->fail('An exception should have been raised');
+            self::$DI['client']->request('GET', '/include/minify/?g=prout');
+        } catch (\Exception $e) {
+        }
     }
 
     public function testFileMinificationWithUnknownFileShouldReturnA500()
     {
-        self::$DI['client']->request('GET', '/include/minify/?f=prout');
-        $this->assertEquals(500, self::$DI['client']->getResponse()->getStatusCode());
+        // there is an exception thrown
+        try {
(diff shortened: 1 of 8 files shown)


tests/Alchemy/Tests/Phrasea/Controller/MinifierTest.php
```diff
@@ -49,13 +49,21 @@ public function testFileMinificationWithoutParamsShouldReturnA400()
 
     public function testFileMinificationWithUnknownGroupShouldReturnA500()
     {
-        self::$DI['client']->request('GET', '/include/minify/?g=prout');
-        $this->assertEquals(500, self::$DI['client']->getResponse()->getStatusCode());
+        // there is an exception thrown
+        try {
+            $this->fail('An exception should have been raised');
```
jygaulier: no. this will jump directly to the catch, so the request is not played.
The request doest NOT throw an exception, but dumps a bunch of " phraseanet.ERROR: Unable to generate data...". Thats is that what we don't want to see during phpunit.

jygaulier: this will jumps directly to the catch, not playing the request.

jygaulier: even if the minifiercontroller throw a 400 or 500, it's catched by symfony http client, and we will always get a valid response. (breakpoint/trace to httpkernel::handle(...catch=true).
It seems this cannot be changed in symfony<3.4 : https://symfony.com/doc/3.4/testing.html#reporting-exceptions.
The test was ok (testing a 500 response).
Error messages should be sent to a log, not polluting the tests output.

jygaulier: The loggers were added by https://github.com/alchemy-fr/Phraseanet/pull/3007/files, so one could think that removing the handlers solves the phpunit pollution issue, but...
- some errors are enforced by a valid test (eg. testing that a bad route returns 5xx). We want the test, but we don't want output.
- some real errors happen, we want to see those - at least to implement a new test that will fail -.

So what we'd like is : 
- a logger to see the real/unwanted errors - that is : any error but the enforced ones.
- a way to avoid the screen dump when the error is "normal" (enforced/covered by a valid test). This "Shut up please" should be inserted on every test before running a "talky" method. 

Any idea ?

btw : disabling the handlers makes phpunit run clean (no text pollution) for __all__ tests.

PR #3260, opened 2019-12-11
PHRAS-2840 : Exclude collection from quarantine checker

aynsix:
### Adds
  - PHRAS-2840: Exclude collection on quarantaine treatment

### DIFF
--- a/lib/Alchemy/Phrasea/Border/Checker/AbstractChecker.php
+++ b/lib/Alchemy/Phrasea/Border/Checker/AbstractChecker.php
@@ -37,7 +37,7 @@ abstract class AbstractChecker implements CheckerInterface
     /**
      * @var \collection[]
      */
-    protected $ignoreCollections = [];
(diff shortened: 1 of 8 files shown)


lib/Alchemy/Phrasea/Border/Checker/AbstractChecker.php
```diff
@@ -34,6 +34,11 @@ abstract class AbstractChecker implements CheckerInterface
      */
     protected $collections = [];
 
+    /**
+     * @var \collection[]
+     */
+    protected $ignoreCollections = [];
```
jygaulier: better name "compareIgnoreCollections" for vars/methods, because "collections" refers to the new record to be checked, while "compare..." refers to the records in db to match against.
btw we may use "ignoreCollections" later as an opposite to "collections", to specify the ok destination(s) as a blacklist or whitelist.

PR #3268, opened 2019-12-13
PHRAS-2255 : add story_total_records on response api/v3/stories

aynsix:
### Adds
  - PHRAS-2255 : in endpoint api/v3/stories , information "story_total_records" is added on the response

### DIFF
--- a/lib/classes/record/adapter.php
+++ b/lib/classes/record/adapter.php
@@ -1820,6 +1820,32 @@ class record_adapter implements RecordInterface, cache_cacheableInterface
         return $this;
     }
 
+    /**
+     * @param  $storyId
+     * @return int
+     * @throws \Exception
+     */
+    public function getStoryTotalRecords($storyId)
+    {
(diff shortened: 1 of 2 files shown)


lib/classes/record/adapter.php
```diff
@@ -1820,6 +1820,32 @@ public function removeChild(\record_adapter $record)
...
+        if (!is_int($storyId)) {
+            throw new \Exception('Only integer for the storyId');
+        }
+
+        $sql = "SELECT count(*) as nb
+                FROM regroup s
+                INNER JOIN record r ON r.record_id = s.rid_child
+                WHERE (s.rid_parent = :storyId) AND (r.parent_record_id = 0)"
+        ;
+
```
jygaulier, line 1839: no. acl must be applied on children.
depending on user's rights on child collection, some children may be not visible.

nmaillat:
PHRAS-2255 Close in favor of pr #3611 and pr #3613

PR #3277, opened 2019-12-18
Fix composer hash

gjacobjn:
# Fixes
 - Temp fix for composer hash


nmaillat:
close in favor of pr #3276

PR #3301, opened 2020-01-22
PHRAS-2879-Define-facets-order-in-query-result

jygaulier:
OK to test

### Changed
- PHRAS-2879: ~Save all facet settings in conf, order is preserved in admin/search-engine-settings.~ DONE
   ~todo : respect order of the admin form while saving.~ DONE
   ~todo : return facets into query results in this setting order.~ DONE
   ~todo : migration from dbox settings (field struct) to conf~ DONE
   todo : remove field facet setting from admin ux ?
   ~todo : fix tests~ DONE

### readme
- facet rendered with /!\ means "new field, not yet configured" (= new dbox mounted ?). The sign will disappear as soon as the settings are saved.
- facet rendered with /?\ means "facet in config, but unknown in current dboxes/structures" (= a dbox was unmounted ?) - the setting is preserved just in case.
- facet with "#" is tech facet

### DIFF
--- a/lib/Alchemy/Phrasea/Core/Version.php
+++ b/lib/Alchemy/Phrasea/Core/Version.php
@@ -16,7 +16,7 @@ class Version
     /**
      * @var string
      */
-    private $number = '4.1.0-alpha.20a';
+    private $number = '4.1.0-alpha.21a';
 
     /**
      * @var string
--- a/config/configuration.sample.yml
+++ b/config/configuration.sample.yml
@@ -28,7 +28,8 @@ main:
             port: 11211
     search-engine:
         type: phrasea
-        options: []
+        options:
+            facets: []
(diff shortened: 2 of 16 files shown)


lib/Alchemy/Phrasea/Core/Version.php
```diff
@@ -16,7 +16,7 @@ class Version
     /**
      * @var string
      */
-    private $number = '4.1.0-alpha.19a';
+    private $number = '4.1.0-alpha.20a';
```
nmaillat: Please bump to 4.1.0-alpha.21a , because 4.1.0-alpha.20a is already use in pr #3328

jygaulier: done

PR #3318, opened 2020-02-03
PHRAS-2879: Add sortable on Aggregates settings

HRavalomanana:
### Adds
  - PHRAS-2879: Add sortable on Aggregates settings

To do : 
  - The action to save status of new model is to do


HRavalomanana:
i close it in favour of  #3321

PR #3319, opened 2020-02-03
PHRAS-2879 sort facet list, solution 2 - two container

HRavalomanana:
### Adds
  - Add 2 columns
  - sort facets : not aggregated on left and other value on right : to do (need to sort facets depending on selected value)
  -add blue background on technival values : done

### DIFF
--- a/lib/Alchemy/Phrasea/SearchEngine/Elastic/ElasticsearchSettingsFormType.php
+++ b/lib/Alchemy/Phrasea/SearchEngine/Elastic/ElasticsearchSettingsFormType.php
@@ -149,7 +149,7 @@ class ElasticsearchSettingsFormType extends AbstractType
 
         // populate aggs to form
         foreach($aggs as $k=>$agg) {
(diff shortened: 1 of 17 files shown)


lib/Alchemy/Phrasea/SearchEngine/Elastic/ElasticsearchSettingsFormType.php
```diff
@@ -104,15 +105,17 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                     "100 values"     => 100,
                     "all values"     => -1
                 ];
+                /*To do : This is just a test but  the filter should depends on selected choice and not that*/
+               $attr = [  'class' => 'aggregate' ];
```
jygaulier: not a good "if" to test dispatch to left or right since only 1 hardcoded facet have "choice" : flashfired).
better : random...

HRavalomanana:
I close it in favour of #3321

PR #3321, opened 2020-02-04
PHRAS-2879 sort facets on aggregates settings

HRavalomanana:
### Adds
  - PHRAS-2879 sort facets on aggregates settings


nmaillat:
close in favor of pr #3301

PR #3337, opened 2020-02-12
PHRAS-2425 allow to generate subdefs in webp.

jygaulier:
### Adds
  - PHRAS-2425: allow to generate subdefs in webp.


nmaillat:
close

PR #3342, opened 2020-02-14
PHRAS-2389 : Phraseanet compliant with strict mode IN MariaDb or Mysql

aynsix:
### Fixes
  - PHRAS-2389 : Phraseanet compliant with strict mode IN MariaDb or Mysql

### DIFF
--- a/lib/conf.d/bases_structure.xml
+++ b/lib/conf.d/bases_structure.xml
@@ -123,7 +123,7 @@
             <type>DATETIME</type>
             <null></null>
             <extra></extra>
-            <default></default>
+            <default>0000-00-00 00:00:00</default>
             <comment></comment>
           </field>
           <field>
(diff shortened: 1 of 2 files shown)


lib/conf.d/bases_structure.xml
```diff
@@ -123,15 +123,15 @@
             <type>DATETIME</type>
             <null></null>
             <extra></extra>
-            <default></default>
+            <default>0000-00-00 00:00:00</default>
```
jygaulier: no invalid dates in strict mode, null is ok if col. is nullable.
https://www.percona.com/blog/2016/10/18/upgrading-to-mysql-5-7-beware-of-the-new-strict-mode/

jygaulier:
last changes ok to me

PR #3365, opened 2020-02-20
PHRAS-2936_write-uuid-before-make-subdef_4.1

jygaulier:
WIP DO NOT MERGE, OK TO PULL FOR TESTING

### Fixes
  - PHRAS-2936: - optims to limit the number of r/w (exiftool) on original file
  - write the uuid before creating record (which raises the make-subdef flag), to prevent collision if the uuid is written while the file is read by make-subdef task.

todo : cleanup debug comments


nmaillat:
Won't do like this close

PR #3439, opened 2020-04-09
PHRAS-3023_slow-query-while-indexing_4.1

jygaulier:
### Fixes
  - fix bound limit depending on order/direction options
  - add index on dbox:record.moddate


jygaulier:
oups pushed too many files

PR #3470, opened 2020-05-07
PHRAS-2865 : add table worker in ab box

aynsix:
### Adds
  - PHRAS-2865:  add table worker in ab box


nmaillat:
include in PR #3481

PR #3482, opened 2020-05-13
PHRAS-3051 Prod - video tools - define a thumbnail for video don't update other static with correct size

HRavalomanana:
### Changed
  - phraseanet-production-client@0.34.199-d

### Fixes
  - PHRAS-3051 Prod - video tools - define a thumbnail for video don't update other static with correct size


nmaillat:
close in favor of #3484 this is only in Phraseanet-production-client

PR #3490, opened 2020-05-19
PHRAS-3078 prod - redirect user on home when received 403 with "end session "

HRavalomanana:
### Fixes
  - prod - redirect user on home when received 403 with "end session "


templates/web/common/menubar.html.twig
```diff
@@ -429,4 +429,36 @@
...
+            },
+            success: function (data) {
+                if (data) {
+                    commonModule.manageSession(data);
+                }
+                var t = 120000;
+                if (data.apps && parseInt(data.apps) > 1) {
+                        t = Math.round((Math.sqrt(parseInt(data.apps) - 1) * 1.3 * 120000));
+                }
+                window.setTimeout("pollNotifications();", t);
```
jygaulier, line 457: let's remove this sh... with t, use 10000

jygaulier:
Not easy to catch every 403 everywhere to alert immediately (we could wrap jquery get, post etc ?). Not now.
to be tested : The "ok" button on the dlg "you must reconnect" does nothing ?

HRavalomanana:
Yes, on click on this button, the user will be redirect at the login page : /login/?redirect=../prod/  (for example)

PR #3496, opened 2020-05-22
PHRAS-3084 Admin - User - users edit  all read right and all admin right in one click

HRavalomanana:
### Adds
  - PHRAS-3084 Admin - User - users edit  all read right and all admin right in one click


templates/web/admin/editusers.html.twig
```diff
@@ -1099,6 +1108,36 @@
...
+      $(this).closest('.right-items').find('.case_right_access div').show();
+      $(this).closest('.right-items').find('input').val(0);
+      $(this).closest('.right-items').find('.select-all-bdd').removeClass('hide');
+      $(this).addClass('hide');
+    });
+
+   function countCheckedRight() {
+     $(".right-items" ).each(function() {
+       let count_checked = $( this ).find('input.checked').length;
+       if ( count_checked == 17) {
```
jygaulier, line 1134: better to compare number of ckboxes (N) with number of checked (n1) and/or unchecked (n0)

PR #3506, opened 2020-05-27
PHRAS-3088 : fix mysql server gone away and dead lock on table worker

aynsix:
## Changelog

- PHRAS-3088 : fix mysql server gone away and dead lock on table worker
- if mysql server is gone away , reconnect it

### DIFF
--- a/lib/Alchemy/Phrasea/WorkerManager/Worker/SubdefCreationWorker.php
+++ b/lib/Alchemy/Phrasea/WorkerManager/Worker/SubdefCreationWorker.php
@@ -81,6 +81,7 @@ class SubdefCreationWorker implements WorkerInterface
 
                 // tell that a file is in used to create subdef
                 $em = $this->repoWorker->getEntityManager();
+                $this->repoWorker->reconnect();
                 $em->beginTransaction();
 
                 try {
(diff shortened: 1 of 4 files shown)


lib/Alchemy/Phrasea/WorkerManager/Worker/SubdefCreationWorker.php
```diff
@@ -170,15 +171,26 @@ public function process(array $payload)
...
+                $em->getConnection()->beginTransaction();
                 try {
                     $workerRunningJob->setStatus(WorkerRunningJob::FINISHED);
                     $workerRunningJob->setFinished(new \DateTime('now'));
                     $em->persist($workerRunningJob);
                     $em->flush();
                     $em->commit();
                 } catch (\Exception $e) {
-                    $em->rollback();
+                    try {
```
jygaulier, line 183: i'm not fan of transaction inside transaction. better ? (to be tested to see if the bug is still fixed)
```
for(ntry=0; ntry<10; ntry++) {
  begin_trans
  try {
     do_trick
     commit
     break
  } catch {
     rollbk
     sleep 1
  }
}
if( ntry == 10) {
  shit alors !
}
```

PR #3541, opened 2020-06-22
PHRAS-3148 Thesaurus GUI - redirect all popup in modal on main windows

HRavalomanana:
### Adds
  - PHRAS-3148 Thesaurus GUI - redirect all popup in modal on main windows


templates/web/thesaurus/new-term.html.twig
```diff
@@ -56,25 +56,18 @@
             </form>
         </center>
         {% else %}
-                {% if nb_candidates_bad > 0 %}
+                {% if nb_candidates_bad > 0 %}t
```
jygaulier, line 59: t ?

templates/web/thesaurus/thesaurus.html.twig
```diff
@@ -315,6 +323,29 @@
 
         <script type="text/javascript">
 
+            //Ajax function to get data
```
jygaulier, line 326: cool but check there is no multi-modal case in app, e.g. a (level 2) confirm dialog opened by a (level 1) edit dialog (to be checked...)
If there is such a case (?), solution : pass "#NEWSY_DLG..." as an arg to function.

PR #3542, opened 2020-06-23
PHRAS-3147 : Thesaurus GUI - launch a populate directly from GUI

aynsix:
### Adds
  - PHRAS-3147 : Thesaurus GUI - launch a populate directly from GUI


jygaulier:
ok pour moi

PR #3556, opened 2020-07-03
PHRAS-2504 : Ginger service integration Auto-subtitle

aynsix:
### Adds
  - PHRAS-2504: Initiate ginga subtitle
  - add main-queue, subtitle-queue
  - add WorkerJob table


lib/Alchemy/Phrasea/WorkerManager/Worker/MainQueueWorker.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public function process(array $payload)
+    {
+
+        // if needed do treatement here depending on the type
+        $queue = null;
+        $messageType = '';
+
+        switch ($payload['type']) {
```
jygaulier, line 30: we could be more generic by avoiding "switch" : the Q name could be directly translated from the payload type.
That way we don't have to touch the "mainQ" code when a new type is added.
To see / call w. Aina :  
- test if a Q exists to reject bad type ? should not happen... 
- create the Q automatically from the type ?

nmaillat:
PHRAS-3081 end of issue, replace sqlite by table in applicationbox for populate and for Phraseanet-service uploader.

PR #3563, opened 2020-07-07
PHRAS-3174_api-v3-records_4.1

jygaulier:
WIP DO NOT MERGE

### fix
- PHRAS-2443 api/v3/stories  number of returned records is  correct

### Adds
- route GET:records
- route POST:records (can create a record without document)
- json can be passed in body OR in a multiparm parm named "body" (to allow upload of file along with json)


jygaulier:
don't forget log, pagination
see ps log service

PR #3568, opened 2020-07-09
PHRAS-3164 Record > export : add tag fields on emails field

HRavalomanana:
### Adds
  -PHRAS-3164 Record > export : add tag fields on emails field

NB :   / or ; or space are replaced by comma on change of the emails field and will be used as separator of each email


templates/web/common/dialog_export.html.twig
```diff
@@ -178,9 +178,10 @@
                         <div>
                             <span style="min-width: 21px;display: inline-block; ">
                                 {{ 'export::mail: destinataire' | trans }}</span>
-                            <input type="text" value="" name="destmail" class="required span4">
+                            <input type="text" value="" class="span4 tagsinput" data-role="tagsinput" >
+                            <input type="hidden" value="" name="destmail" class="required span4 destmail" >
```
jygaulier, line 182: could we use the taginput directly without converting to destmail ? We can change the back to accept ',' or ';', ' ', ... as separators.

HRavalomanana: Ok noted, in that case we need to use `,`  as separators on back

HRavalomanana:
Added : autocomplete on email field

PR #3576, opened 2020-07-23
PHRAS-3181 : Worker video build subdefinition ended with error and tmp file is not copied to destination

aynsix:
## Changelog

- add error status on WorkerRunningJob to have traceability of a build
- add column info  to save actually number of  attempt if there is error
- display info on hover


nmaillat:
PHRAS-3181 merge 
add error status on WorkerRunningJob to have traceability of a build
add column info to save actually number of attempt if there is error
display info on hover

Issue #3641, opened 2020-11-06
Upgrade from 3.8.8 > 4.0 > 4.1: Scheduler does not start


frague59:
Hi, 
I'm finishing an upgrade a Phraseanet 3.8.8 to 4.1, using an intermediate 4.0, from debian 8 to debian 9. I'll stay on a debian 9 because of php 7.0 not available in debian 10.

I've almost finished my upgrade, the application front works fine, but the scheduler fails to start.

[scheduler-2020-11-06.log](https://github.com/alchemy-fr/Phraseanet/files/5501408/scheduler-2020-11-06.log)

It seems the problems come from the lack of a `jeton` field somewhere.

I have an 500 error in the Admin > Task manager in front too - when I try to start the scheduller.

Have you any idea on how to fix it ? 

Thanks for this great application !

frague59:
It seems the indexing task is faulty.

frague59:
This error appears in Phraseanet 4.1, on a debian 9 server.

```sh
$ php --version 
PHP 7.0.33-0+deb9u10 (cli) (built: Oct  6 2020 17:08:28) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.33-0+deb9u10, Copyright (c) 1999-2017, by Zend Technologies
```
```sh
# ./bin/console -V

  _____  _    _ _____            _____ ______          _   _ ______ _______
 |  __ \| |  | |  __ \    /\    / ____|  ____|   /\   | \ | |  ____|__   __|
 | |__) | |__| | |__) |  /  \  | (___ | |__     /  \  |  \| | |__     | |
 |  ___/|  __  |  _  /  / /\ \  \___ \|  __|   / /\ \ | . ` |  __|    | |
 | |    | |  | | | \ \ / ____ \ ____) | |____ / ____ \| |\  | |____   | |
 |_|    |_|  |_|_|  \_|_/    \_\_____/|______/_/    \_\_| \_|______|  |_|
  _  __                 _       _  __                             _
 | |/ /___ _ _  ___ ___| |___  | |/ /___ _ __  _ __  __ _ _ _  __| |___ _ _
 | ' </ _ \ ' \(_-</ _ \ / -_) | ' </ _ \ '  \| '  \/ _` | ' \/ _` / -_) '_|
[...]

PR #3643, opened 2020-11-09
Fix indexer start SQL errorrs

frague59:
### Changed
  - Updates SQL queries in Fetcher.php and ScheduledFetcherDelegate.php
  
### Fixes
  - The task manager did not start because of SQL error in the indexer (#3641)


nmaillat:
Indexation task is not used anymore

PR #3656, opened 2020-11-19
PHRAS- 3274 Prod - editing - request is to long when editing a large amount of records

aynsix:
### Fixes
  - PHRAS-3274: Prod - editing - request is to long when editing a large amount of records

### DIFF
--- a/lib/Alchemy/Phrasea/Controller/Prod/StoryController.php
+++ b/lib/Alchemy/Phrasea/Controller/Prod/StoryController.php
@@ -73,12 +73,6 @@ class StoryController extends Controller
         }
 
         $recordAdapter = $story->set_metadatas($metadatas);
-        // order to write meta in file
-        $this->dispatch(WorkerEvents::RECORDS_WRITE_META,
-            new RecordsWriteMetaEvent([$story->getRecordId()], $story->getDataboxId()));
-
-        // tell phraseanet to rebuild subdef
-        $this->dispatch(RecordEvents::SUBDEFINITION_CREATE, new SubdefinitionCreateEvent($recordAdapter));
 
         $storyWZ = new StoryWZ();
         $storyWZ->setUser($this->getAuthenticatedUser());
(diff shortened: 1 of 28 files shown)


jygaulier:
I had another idea :
Put the whole original request as payload, then the worker simply has to play this request again.
That way the original controller is left untouched, and if we modify the editing later (front), we can change the controller (natural way), without having to dig in the worker's code.

To avoid endless loop, we can :
- change the "ok" buton route to a tiny controller (build event, send)
or (prefered)
- keep a single route, but add a parameter to the replayed request to say "do the job"

lib/Alchemy/Phrasea/Controller/Prod/StoryController.php
```diff
@@ -71,6 +73,10 @@ public function postCreateFormAction(Request $request)
         }
 
         $recordAdapter = $story->set_metadatas($metadatas);
+        // order to write meta in file
+        $this->dispatch(WorkerEvents::RECORDS_WRITE_META,
+            new RecordsWriteMetaEvent([$story->getRecordId()], $story->getDataboxId()));
+
         // tell phraseanet to rebuild subdef
         $this->dispatch(RecordEvents::SUBDEFINITION_CREATE, new SubdefinitionCreateEvent($recordAdapter));
```
jygaulier, line 82: to be verified : we may not need to create subdefs for a story, since it has no 'document'. Check that the "folder" thumbnail / preview are still displayed.

lib/Alchemy/Phrasea/WorkerManager/Subscriber/RecordSubscriber.php
```diff
@@ -142,59 +144,60 @@ public function onRecordCreated(RecordEvent $event)
...
+                    } else {
+                        $payload = [
+                            'message_type' => MessagePublisher::WRITE_METADATAS_TYPE,
+                            'payload' => [
+                                'recordId'      => $recordId,
+                                'databoxId'     => $databoxId,
+                                'subdefName'    => $subdef->get_name()
+                            ]
+                        ];
+
```
jygaulier, line 183: payload is same, can be factorized

jygaulier:
I pull and test

lib/Alchemy/Phrasea/WorkerManager/Subscriber/RecordSubscriber.php
aynsix: ok

PR #3658, opened 2020-11-19
PHRAS-3279_api-v3-permalinks-to-es_4.1

jygaulier:
## Changelog
The subdefintion's Permalinks are now injected in Elasticsearch index when the new option  "populate_permalinks" is set to "True".
This option is define in configuration.yml, "search-engine" section.
If property is missing the value is false.
With this feature, "api/v3/searchraw" can serve subdefinition's permalinks.
Warning, when populate_permalinks:true, duration and memory consumption during the Populate are increase, this factor is depending of  "databox's" structure.

### Adds
  - PHRAS-3279 : permalinks  in Elasticsearch index.

### DIFF
--- a/lib/Alchemy/Phrasea/SearchEngine/Elastic/Indexer/Record/Hydrator/SubDefinitionHydrator.php
+++ b/lib/Alchemy/Phrasea/SearchEngine/Elastic/Indexer/Record/Hydrator/SubDefinitionHydrator.php
@@ -32,55 +32,38 @@ class SubDefinitionHydrator implements HydratorInterface
 
     public function hydrateRecords(array &$records)
     {
-        $sql = <<<SQL
-            SELECT
-              s.record_id,
-              s.name,
-              s.height,
-              s.width,
-              CONCAT(TRIM(TRAILING '/' FROM s.path), '/', s.file) AS path
-            FROM subdef s
-            WHERE s.record_id IN (?)
-            ORDER BY s.record_id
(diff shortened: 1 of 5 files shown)


lib/Alchemy/Phrasea/SearchEngine/Elastic/Indexer/Record/Hydrator/SubDefinitionHydrator.php
```diff
@@ -11,76 +11,66 @@
...
-            WHERE s.record_id IN (?)
-            ORDER BY s.record_id
-SQL;
-        $statement = $this->databox->get_connection()->executeQuery($sql,
-            array(array_keys($records)),
-            array(Connection::PARAM_INT_ARRAY)
-        );
+        foreach(array_keys($records) as $rid) {
+            try {
+                $subdefs = $this->databox->getRecordRepository()->find($rid)->get_subdefs();
```
jygaulier, line 41: this is the sloooow point.
We should include this into the "if", but that means restoring most of the previous code (sql) into a big "else".
So for readability, for now the best is : 

```
pub fct hydrateRecords(...)
{
  if(test) hydrateRecordsWithPermalinks(...) else hydrateRecordsWithoutPermalink(...)
}
priv hydrateRecordsWithPermalinks(...) { new code };
priv hydrateRecordsWithoutPermalinks(...) { old code };
```

PR #3669, opened 2020-12-09
PHRAS-3278 Release version 4.1.3  Migration patch for 4.1.3

aynsix:
### Adds
  - PHRAS-3278: Migration patch for 4.1.3

requires simultaneous fusion of PR

 - #3680 Refacto worker queue name and declaration
 - #3663 invert "push" and "feedback" in timeline require a migration of log


lib/classes/patch/413.php
```diff
@@ -0,0 +1,100 @@
...
+        $stmt->closeCursor();
+
+        // remove temp action
+        $sql = "ALTER TABLE log_docs CHANGE action action ENUM('push','add','validate','edit','collection','status','print','substit','publish','download','mail','ftp','delete','') CHARACTER SET ascii COLLATE ascii_bin NOT NULL";
+        $stmt = $databox->get_connection()->prepare($sql);
+        $stmt->execute();
+        $stmt->closeCursor();
+
+        // patch for reminder validation key, default value to 20
+
```
jygaulier, line 80: warning ! because this patch applies on databoxes, the following code will run many times.
Find how to apply a patch on appbox and dboxes

PR #3689, opened 2021-01-13
PHRAS-3330 add worker queue name env and allow worker run exploded

moctardiouf:
### Adds
  - PHRAS-3330: Add Worker queue-name env
  - Enable running queues individualy with different allowed sub process number


moctardiouf:
I need to add a test on available number of subpresses and break when one is missing

PR #3725, opened 2021-02-16
PHRAS-3080 When a document is exported , add a PDF File with CGU

aynsix:
### Adds
  - PHRAS-3080:  When a document is exported , add a PDF File with CGU


jygaulier:
ok for me

PR #3749, opened 2021-03-16
PHRAS-3392 Worker recordEdit - apply description made by Tx as classification PHRAS-3381

aynsix:
### Adds
  - PHRAS-3392: Worker recordEdit - apply description made by Tx as classification PHRAS-3381


lib/Alchemy/Phrasea/Controller/Prod/EditController.php
```diff
@@ -353,16 +353,14 @@ public function applyAction(Request $request) {
      */
     public function applyJSAction(Request $request): JsonResponse
     {
-        // todo : by worker
-
-        // for now call record_adapter. no check, no acl, ...
         /** @var stdClass $arg */
         $arg = json_decode($request->getContent());
+        $databoxId = reset($arg->records)->sbas_id;
```
jygaulier, line 358: ~~no. even if we "know" that sbas_id is always the same, this may change later.
since "records" is part of body, we should pass null as last arg.~~
correction :
It seems that the existing worker code relies heavily on a unique databox...
So we must change the json to receive a unique sbid, and an array of rids.

jygaulier (changes requested):

lib/Alchemy/Phrasea/Controller/Prod/EditController.php
aynsix: ok , I've change that

lib/Alchemy/Phrasea/WorkerManager/Worker/RecordEditWorker.php
```diff
@@ -30,122 +31,130 @@ public function __construct(WorkerRunningJobRepository $repoWorker, EventDispatc
...
-                $em->commit();
-            } catch (\Exception $e) {
-                $em->rollback();
+        } else {
+            $arg = json_decode($payload['data']);
+            $recordIds = array_column($arg->records, 'record_id');
+
+            // for now call record_adapter. no check, no acl, ...
+            foreach($arg->records as $rec) {
+                $r = $databox->get_record($rec->record_id);
```
jygaulier, line 137: add try/catch, test r=null (like for mds)
the record could have been deleted before worker.

PR #3799, opened 2021-05-20
PHRAS-3442_optimize-list-notifications_4.1-bis

jygaulier:
### Fixes
  - PHRAS-3442: pull notifications every 60 secs.


jygaulier:
we must see why there is 2 poll on the same route, and if possible 
- merge code to have only one.
or if there is a good reason for mutiple poll (one fot bar/notifs ? one for received baskets ?)
- add an argument so we don't do the same sql

jygaulier:
todo :
- The 'eye" button (mark a notification as "seen") in the dropdown does not work. see `resources/www/common/js/components/common.js::142 `; It works in dlg, see `Phraseanet-production-client/src/components/notify/notifyLayout.js::197`
To be implemented : In dropdown, after marking a notif as read, the content must be updated (because **the dropdown shows only unread notifs** - while **the dlg shows both unread an read**) 

- Deleting notifs 24h after reading is commented during test ; see `lib/classes/eventsmanager/broker.php::239`.
Deletion is based on creation datetime (bad) because there is not yet "read" datetime. To be implemented as read "read_on" with null when not read --> no more "unread" flag :) (but requires a migration).

- route `session/update` is not used anymore, but is kept for unit-tests. Check the everything that was done (eg. changing user) is implemented in `user/notifications`. re-using this route is maybe a good idea for semantic reasons (why "notification" should keep a "session" ?)

- test and remove commented old code...

PR #3811, opened 2021-06-16
PHRAS-3438 : conf/configuration.yml - Set an http and ftp proxy and use serveur Side eg worker ; geocoding request, ftp etc...

aynsix:
### Adds
  - PHRAS-3438: conf/configuration.yml - Set an http and ftp proxy and use serveur Side eg worker ; geocoding request, ftp etc...
 
 - configuration
 `network-proxies:`
        `http-proxy:`
        `enabled:` `true`
        `host:`
        `port:`
        `user:`
       ` password:`

 -  PHRAS-3464  launch PMA, ELK, FTP containers with their own docker-compose file, for more compositing
  `docker-compose down` is recommended before deploy this PR , check also our COMPOSE_FILE for selecting which containers will be launch.


jygaulier:
seems ok.
todo : 
- code/test with all options : auth (?)
- apply the "NO_PROXY" rule (?) ;  may be not, since it may be handled by the proxy itself ?

PR #3815, opened 2021-06-24
PHRAS-3468_comment-traces

jygaulier:
### Fixes
  - PHRAS-3468: commented logs / direct write to "logs/traces.txt" because php7 syntax prevents strings extraction.


jygaulier:
closed in favor of #3828

PR #3823, opened 2021-07-01
PHRAS-3474 expose field setting

aynsix:
### Adds
  - PHRAS-3474: expose field setting


lib/Alchemy/Phrasea/PhraseanetService/Controller/PSExposeController.php
```diff
@@ -711,6 +713,75 @@ public function addPublicationAssetsAction(PhraseaApplication $app, Request $req
...
+    /**
+     * @param PhraseaApplication $app
+     * @param Request $request
+     * @return \Symfony\Component\HttpFoundation\JsonResponse
+     */
+    public function saveFieldMappingAction(PhraseaApplication $app, Request $request)
+    {
+        $exposeName = $request->get('exposeName');
+        $fields = ($request->request->get('fields') === null) ? null : json_encode(array_keys($request->request->get('fields')));
+
```
jygaulier, line 760: too easy !
the goal is to save settings to an extensible format.
If tomorrow we want to add attributes to a field (eg. a simple color ore a complex acl), event in this json it's not possible because it's a simple list)
we should
- save each field as a specific setting (child of a common "fields" parent), with fieldname/or/int as key.
or
- save all fields as json, but with each field as an (extensible) object

aynsix:
to be closed , mapping is saved now in profile #3827

PR #3845, opened 2021-08-12
PHRAS-3348 gracefull delay of 30 days before removing expired validation token

aynsix:
### Fixes
  - PHRAS-3348: gracefull delay of 30 days before removing expired validation token


jygaulier:
ok to me

PR #3858, opened 2021-09-09
PHRAS-3522 Prod - Notification - notifications windows contain uninterpreted HTML (URL)

aynsix:
### Fixes
  - PHRAS-3522: Prod - Notification - notifications windows contain uninterpreted HTML (URL)


jygaulier:
ok to me

PR #3874, opened 2021-10-01
PHRAS-3540 Some export email stay in "running" status

aynsix:
### Fixes
  - PHRAS-3540: Some export email stay in "running" status


jygaulier:
ok to me

PR #3879, opened 2021-10-05
PHRAS-3541 prod - Image rotation NOK - ETAG is not renew - File is correctly rotated but not invalidate in browser cache.

aynsix:
### Fixes
  - PHRAS-3541: prod - Image rotation NOK - ETAG is not renew - File is correctly rotated but not invalidate in browser cache.

### DIFF
--- a/lib/Alchemy/Phrasea/Twig/PhraseanetExtension.php
+++ b/lib/Alchemy/Phrasea/Twig/PhraseanetExtension.php
@@ -343,7 +343,7 @@ class PhraseanetExtension extends \Twig_Extension
         } elseif ($record instanceof \record_adapter) {
             if (null !== $thumbnail = $record->get_subdef($subdefName)) {
                 if ('' !== $path = $thumbnail->getRealPath()) {
-                    $etag = $thumbnail->getEtag();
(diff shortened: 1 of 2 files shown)


lib/classes/media/subdef.php
```diff
@@ -573,6 +574,10 @@ public function rotate($angle, Alchemyst $alchemyst, MediaVorus $mediavorus)
         $this->width = $media->getWidth();
         $this->height = $media->getHeight();
 
+        // generate a new etag after rotation
```
jygaulier, line 577: simple `$this->etag = null;` should enforce creation on next getEtag();

lib/classes/media/subdef.php
```diff
@@ -362,11 +362,12 @@ public function get_record_id()
 
     public function getEtag()
     {
-        if (!$this->etag && $this->is_physically_present()) {
-            $file = new SplFileInfo($this->getRealPath());
-
+        $file = new SplFileInfo($this->getRealPath());
+        // etag is generated from md5 of path and file last modified time
+        // so check if etag is to be changed
+        if ((!$this->etag && $this->is_physically_present()) || $this->etag != md5($file->getRealPath() . $file->getMTime())) {
```
jygaulier, line 368: `md5($file->getRealPath() . $file->getMTime()) `is the same value as `generateEtag()`
so in fact we do `if(x != f()) then x=f()` --> `x=f()` if faster

PR #3929, opened 2021-12-14
PHRAS-3602_shared-baskets-step1 WIP DO NOT MERGE

jygaulier:
WIP DO NOT MERGE

### Changed (should not bc-break)
  - move "validation" tables and code to "basket"
  - bump to 4.1.6
  - new tables ab.BasketParticipants ; ab.BasketElementVotes
  - todo : remove ab.validation*

PR #3998, opened 2022-03-28
PHRAS-3590 Prod -detailed view - feedback tab - change user right require for seing the tab content

aynsix:
### Adds
  - PHRAS-3590: Prod -detailed view - feedback tab - change user right require for seing the tab content


nmaillat:
PHRAS-3590 moved tp PR #3929

PR #4006, opened 2022-04-08
PHRAS-3662 Prod - shared basket - fix icon in detailed view and action bar

aynsix:
### Fixes
  - PHRAS-3662: Prod - shared basket - fix icon in detailed view and action bar


jygaulier:
some basket can show 2 icons on workzone, but only 1 is visible in preview.
here a basket was first received from push, then shared.
image
image
image

PR #4007, opened 2022-04-19
PHRAS-3662 Share Basket fix GUI - icons color and behavior

aynsix:
### Fixes
  - PHRAS-3662: fix some display icon
  - fix the icon on right on share
  - separate feedback option with string
  - inverse push and share icon in action bar
  - fix multi button , extend white background


Phraseanet-production-client/src/components/record/pushOrShare/pushOrShare.js
```diff
@@ -130,6 +130,44 @@ const pushOrShare = function (services, container) {
...
+            var that = $(this);
+            if($("INPUT[name=isFeedback]").val() === '0') {
+                // just a share
+                that.css({'background-color':'#fff', 'color': '#4caf50'});
+            } else {
+                that.css({'background-color': '#fff', 'color': '#E0215D'});
+            }
+        }
+    );
+
```
jygaulier, line 151: no hardcoded colors in .js, only in .scss ; use color aliases, harcode theme color only once (except for "standard" colors like black/white).
in js, add-or-remove-or-change a class.

PR #4009, opened 2022-04-28
PHRAS-3668 Add autosub exec to phraseanet worker

moctardiouf:
### Adds
  - PHRAS-3668: add autosub exec to phraseanet worker
  - implementation of automatic subtitle generating using google api


nmaillat:
close in favor of #4246

PR #4021, opened 2022-05-18
PHRAS-3675 fix hearbeat

aynsix:
### Fixes
  - PHRAS-3675: fix hearbeat


jygaulier:
seems ok : hearbeat in the most external loop on the processes pool, while waiting any on the async process to end.

PR #4023, opened 2022-05-18
PHRAS-3678: LightBox  fix popup when vote

aynsix:
### Fixes
  - PHRAS-3678: LightBox  fix popup when vote


jygaulier:
tested but strange:
- without the fix, the message "do you want to send" appears each vote
- with the fix, a message appears only when all votes are done, but the message is empty ???
image

PR #4024, opened 2022-05-19
PHRAS-3664 Prod - record moving between collections action : no retry on indexation failure.

aynsix:
### Fixes
  - PHRAS-3664: Prod - record moving between collections action : no retry on indexation failure.


lib/classes/record/adapter.php
```diff
@@ -566,6 +566,8 @@ public function get_duration()
     public function move_to_collection(collection $collection, appbox $appbox = null)
     {
         if ($this->getCollection()->get_base_id() === $collection->get_base_id()) {
+            $this->dispatch(RecordEvents::COLLECTION_CHANGED, new CollectionChangedEvent($this, $collection, $collection));
+
```
jygaulier, line 570: possibly bad side effects with code (plugins ?) that listen to **real** collection changes.
we must fix the (lost) indexation pb where it happens (e.g. posting a reindex evt)

PR #4036, opened 2022-05-31
PHRAS-3687: Remove .env non-closed double-quot that breaks design in html doc

felixalchemy:
### Fixes
  - PHRAS-3687: Remove .env non-closed double-quot that breaks design in html doc


nmaillat:
include in pr #4038

PR #4038, opened 2022-06-01
PHRAS-3663 persist state workzone filter

aynsix:
### Fixes
  - PHRAS-3663 persist state workzone filter in user setting

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,7 @@
 version: 2.1
 orbs:
-    aws-ecr: circleci/aws-ecr@4.0.1
+    aws-ecr: circleci/aws-ecr@6.15.0
+    aws-cli: circleci/aws-cli@1.3.1
 executors:
   docker_build:
     machine:
@@ -105,7 +106,7 @@ jobs:
 
     build_phraseanet-fpm:
         machine:
-          image: ubuntu-1604:201903-01
+          image: ubuntu-2004:202201-02
           docker_layer_caching: true
         working_directory: ~/alchemy-fr/Phraseanet
         steps:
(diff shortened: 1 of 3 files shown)


templates/web/prod/WorkZone/Macros.html.twig
```diff
@@ -336,7 +356,11 @@
                              class="ui-accordion-header ui-state-default
                              ui-corner-all header SSTT {% if not basket.getId() %}unread{% endif %}
                              {% if basket.getId() == selected_id and selected_type == 'basket' %}active{% endif %}
-                             {{ classes }}">
+                             {{ classes }}
+                             {# block defined in lib/Alchemy/Phrasea/Helper/WorkZone.php #}
```
jygaulier, line 360: do not duplicate responsibility. 
- twig restore ux and initiate checkboxes as chkd / unchkd depending on saved preferences. twig doesn't know the effect of ckboxes.
- js display or hide elements.

Phraseanet-production-client/src/components/ui/workzone/index.js
```diff
@@ -482,6 +482,18 @@ const workzone = (services) => {
...
+                data: {
+                    prop: $(this).attr("data-prop"),
+                    value: $(this).is(':checked') ? 1 : 0
+                },
+                success: function (data) {
+                    return;
+                }
+            });
         });
     }
```
jygaulier, line 498: During init, show / hide blocks depending on checkboxes. loop on $('#basket-filter INPUT')
This ensure we use the same algorithm on init / change. No risk to forget to replicate a js effect to the twig.

PR #4055, opened 2022-06-10
PHRAS-3525 Admin can change status to canceled for multiple job

aynsix:
### Adds
  - PHRAS-3525: Admin can change status to canceled for multiple job


jygaulier:
ok to me

PR #4063, opened 2022-06-17
PHRAS-3697 printed pdf improvement

aynsix:
### Fixes
  - PHRAS-3697: printed pdf improvement
  - show/hide record information  on preview
  - add column "printable" in metadata_structure / patch version 4.1.6
  - can change font size of description


jygaulier:
ok to me

PR #4097, opened 2022-07-13
PHRAS-3704 add info to subdef create webhook payload

4rthem:
@aynsix to be completed :)
Thanks!

## TODO

- [ ] get `verify` (verify SSL) option from configuration
- [ ] inject instance hostname in URL generator


nmaillat:
PHRAS-3704 close in favor of #4098

PR #4104, opened 2022-07-21
PHRAS-3712 admin - sudefinition - property to define building rules on subdefinition

aynsix:
### Adds
  - PHRAS-3712: admin - sudefinition - property to define building rules on subdefinition
  - add atribute "tobuild" default value true

### DIFF
--- a/lib/Alchemy/Phrasea/Command/BuildSubdefs.php
+++ b/lib/Alchemy/Phrasea/Command/BuildSubdefs.php
@@ -406,11 +406,6 @@ class BuildSubdefs extends Command
                         continue;
                     }
 
-                    if (!$subdef->getDataboxSubdef()->isTobuild()) {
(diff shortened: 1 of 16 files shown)


lib/Alchemy/Phrasea/Command/BuildSubdefs.php
```diff
@@ -406,11 +406,6 @@ protected function doExecute(InputInterface $input, OutputInterface $output)
                         continue;
                     }
 
-                    if (!$subdef->getDataboxSubdef()->isTobuild()) {
-                        unset($subdefNamesToDo[$name]);
-                        continue;
-                    }
-
                     if(!in_array($name, $this->subdefsByType[$type])) {
```
jygaulier, line 409: add OR test to prune subdefs that are defined in struct, but with tobuild=false

PR #4114, opened 2022-08-10
PHRAS-3729 add extra_hosts to worker to support uploader ingestion

4rthem:


nmaillat:
close in favor of #4116

PR #4115, opened 2022-08-23
PHRAS-3727 Prod - shared basket - Quit share is not allowed

aynsix:
### Fixes
  - PHRAS-3727: Prod - shared basket - Quit share is not allowed


jygaulier:
ok to me

PR #4118, opened 2022-09-06
PHRAS-3719 filter on job type

aynsix:
### Fixes
  - PHRAS-3719: filter on job type


templates/web/admin/worker-manager/worker_info.html.twig
```diff
@@ -135,19 +145,25 @@
 
             $.ajax({
                 type: "GET",
-                url: "/admin/worker-manager/info?reload=1&running="+running+"&finished="+finished+"&error="+error+"&interrupt="+interrupt,
+                url: "/admin/worker-manager/info?reload=1&running="+running+"&finished="+finished+"&error="+error+"&interrupt="+interrupt+"&jobType="+jobType,
```
jygaulier, line 148: checck if jobtype is url-escaped.
better to use `data` arg

PR #4139, opened 2022-09-26
PHRAS-3694 bin/maintenance clean:users finish

aynsix:
### Fixes
  - PHRAS-3694: bin/maintenance clean:users finish

### DIFF
--- a/lib/Alchemy/Phrasea/Command/Maintenance/CleanUsersCommand.php
+++ b/lib/Alchemy/Phrasea/Command/Maintenance/CleanUsersCommand.php
@@ -76,7 +76,7 @@ class CleanUsersCommand extends Command
                 . "- <info>1 year</info>\n"
                 . "\<USERTYPE>specify type of user to clean : \n"
                 . "- <info>admin</info> \n"
-                . "- <info>appowner</info \n>"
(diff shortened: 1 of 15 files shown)


lib/Alchemy/Phrasea/Command/Maintenance/CleanUsersCommand.php
```diff
@@ -105,13 +123,48 @@ protected function doExecute(InputInterface $input, OutputInterface $output)
...
+            $stmt->closeCursor();
+
+            $helper = $this->getHelper('question');
+            $question = new ConfirmationQuestion(sprintf("Confirm deletion of %s deleted user(s) [y/n] : ", $nDeleted), false);
+
+            if (!$helper->ask($input, $output, $question)) {
+                return 0;
+            }
+
+            $sqlDelete = "DELETE FROM Users WHERE deleted = 1";
```
jygaulier, line 158: normally, a "deleted=1" user should not have related data (basket, account, share, feedback, stories, log...) but we should delete (?) everything to be sure.
Q @nmaillat : purge will delete everything related ! maybe an an option `--new-owner=email` (or id) ?
We could also dynamically create a user "lost+found" ?

PR #4180, opened 2022-11-09
PHRAS-2995_phr-as-auth-idp-BIS

jygaulier:
### Adds
  - PHRAS-2995: auth provider to ps-auth ; **todo: remove debug code**
  - bump to '4.1.7-rc1' with conf/authentication/provider migration
  ```      
authentication:
    ...
    providers:
    ...
        ps_auth_1:
            enabled: false
            display: false
            title: 'PS Auth 1'
            type: 'ps-auth'
            options:
                client-id: ''
                client-secret: ''
                base-url: 'https://api-auth.phrasea.local'
                provider-type: 'oauth'
                provider-name: 'v2'
                icon-uri: null,
                debug: false
                birth-group: 'firstlog'
                everyone-group: '_everyone'
                metamodel: '_metamodele'
                model-gpfx: '_M_'
                model-upfx: '_U_'
                auto-logout: false
```
### Removes
  - PHRAS-2995: removed google+ auth provider

### CODE REVIEW

File: lib/Alchemy/Phrasea/Controller/Root/LoginController.php [hunk 1]
```diff
@@ -688,6 +688,10 @@ public function authenticationCallback(Request $request, $providerId)
         try {
             $provider->onCallback($request);
             $token = $provider->getToken();
+            $options = $provider->getOptions();
+            if (isset($options['base-url'])) {
+                $this->getSession()->set('auth_provider.base-url', $options['base-url']);
+            }
```
jygaulier, line 694: you should not need to set this in session. you can fetch the provider with session('auth_provider.id'), then get options

lib/Alchemy/Phrasea/PhraseanetService/Controller/PSExposeController.php
```diff
@@ -183,6 +184,16 @@ public function listPublicationAction(PhraseaApplication $app, Request $request)
 
         $session = $this->getSession();
         $passSessionName = $this->getPassSessionName($request->get('exposeName'));
+        $providerId = $session->get('auth_provider.id');
+
+        if (!$session->has($passSessionName) && $providerId != null) {
+            try {
+                $provider = $this->getAuthenticationProviders()->get($providerId);
```
jygaulier, line 191: are you sure this provider is of type ps-auth ?

templates/web/prod/WorkZone/Macros.html.twig
```diff
@@ -668,12 +668,16 @@
 {% macro make_expose_bloc(app, WorkZone) %}
     <div id="expose_workzone" class="expose_workzone">
         <div id="expose_sel" class="custom_select_dark" style="padding:10px;">
-            {% set expose_list= app['conf'].get(['phraseanet-service', 'expose-service', 'exposes']) %}
-
+            {% set expose_list = app['conf'].get(['phraseanet-service', 'expose-service', 'exposes']) %}
+            {% set auth_base_url = app['session'].get('auth_provider.base-url') | trim('/') %}
```
jygaulier, line 672: get the provider, then it's type, then its options

lib/Alchemy/Phrasea/Controller/Root/LoginController.php
aynsix: ok

templates/web/prod/WorkZone/Macros.html.twig
aynsix: ok

lib/Alchemy/Phrasea/PhraseanetService/Controller/PSExposeController.php
```diff
@@ -189,8 +189,10 @@ public function listPublicationAction(PhraseaApplication $app, Request $request)
         if (!$session->has($passSessionName) && $providerId != null) {
             try {
                 $provider = $this->getAuthenticationProviders()->get($providerId);
-                $session->set($passSessionName, $session->get($provider->getId() . '.provider.access_token'));
-                $session->set($this->getLoginSessionName($request->get('exposeName')), $session->get($provider->getId() . '.provider.username'));
+                if ($provider->getType() == 'PsAuth') {
+                    $session->set($passSessionName, $session->get($provider->getId() . '.provider.access_token'));
+                    $session->set($this->getLoginSessionName($request->get('exposeName')), $session->get($provider->getId() . '.provider.username'));
+                }
```
jygaulier, line 195: you are not supposed to set/get in session "internal" provider values (prefixed by the provider id), names may change anytime.
you can only get the provider then call getter/setters.
If u need missing infos from the provider we will add it.

PR #4191, opened 2022-12-06
PHRAS-3783 Prod - Order manager - add Download Expiration date

aynsix:
### Adds
  - PHRAS-3783: Prod - Order manager - add Download Expiration date
  - bump version
  - migration patch for table record_right
  -  add config key
  `order-manager:
        download-hd:
             expiration-days: 15`


lib/classes/patch/417RC2.php
```diff
@@ -0,0 +1,77 @@
...
+    private function patch_databox(databox $databox, Application $app)
+    {
+    }
+
+    private function patch_appbox(base $appbox, Application $app)
+    {
+        $cnx = $appbox->get_connection();
+
+        $expirationDays = $app['conf']->get(['order-manager', 'download-hd', 'expiration-days'], "15");
+        $expireOn = (new \DateTime('+ '. $expirationDays .' day'))->format('Y-m-d h:m:s');
```
jygaulier, line 69: danger : existing links wil expire ? expiration should apply only to new links (? check with @nmaillat )

lib/classes/patch/417RC2.php
```diff
@@ -0,0 +1,77 @@
...
+
+    private function patch_databox(databox $databox, Application $app)
+    {
+    }
+
+    private function patch_appbox(base $appbox, Application $app)
+    {
+        $cnx = $appbox->get_connection();
+
+        $expirationDays = $app['conf']->get(['order-manager', 'download-hd', 'expiration-days'], "15");
```
jygaulier, line 68: Since we apply patch, the conf entry _probably_ does not yet exists.
We should add the new conf entry to conf file - if it does not already exists -.
To avoid bc break I suggest null as default, so **existing** rights will not expire.

lib/classes/patch/417RC2.php
jygaulier: ok only "order" case is updated

PR #4208, opened 2023-01-10
PHRAS-3778 deactivate webhook when many times undelivered in the url

aynsix:
### Fixes
  - PHRAS-3778: deactivate webhook when many times undelivered in the url


aynsix:
https://github.com/alchemy-fr/Phraseanet/pull/4209

PR #4219, opened 2023-01-19
PHRAS-3800_xss

jygaulier:
### Fixes
  - PHRAS-3800: add encode option to record::get_title ; render preview.record_title in twig

todo: facet values


aynsix (commented):
ok for me

PR #4239, opened 2023-01-30
PHRAS-3804: cookies consent

aynsix:
### Adds
  - PHRAS-3804: Add cookies consent
  - `main :
         cookie_consent : false`


nmaillat:
close in favor of #4244

PR #4246, opened 2023-02-10
PHRAS-3668 use autosub for subtitle

aynsix:
### Adds
  - PHRAS-3668:  use autosub for subtitle
  - remove ginger


nmaillat:
replace PR #4009

nmaillat:
replace by #4405

PR #4289, opened 2023-03-30
PHRAS-3830 prod - advance search - operator for field type number

aynsix:
### Adds
  - PHRAS-3830: prod - advance search - operator for field type number


jygaulier:
ok to me

PR #4320, opened 2023-05-19
PHRAS-3848 prod - order manager - cancel an order

aynsix:
### Adds
  - PHRAS-3848: prod - order manager - cancel an order


jygaulier:
code ok to me

PR #4343, opened 2023-07-21
PHRAS-3876 Prod - Create - Story Title fix

aynsix:
### Fixes
  - PHRAS-3876: Prod - Create - Story Title fix


jygaulier:
no production-client version bump ?

PR #4366, opened 2023-08-10
PHRAS-3889 Worker - metadata write - mime/type whitelist - write metadatas only on whitelisted files

aynsix:
### Adds
  - PHRAS-3889: Worker - metadata write - mime/type whitelist - write metadatas only on whitelisted files


jygaulier:
ok to merge

jygaulier:
see  https://exiftool.org/ : "Supported File Types"
Lets add commonly used:
- tif
- eps
- pdf

PR #4383, opened 2023-09-26
PHRAS-3921 add client_credential token renewal

4rthem:
### Changed
  - Add support to access token expiration for client_credentials grant type in Expose plugin


nmaillat:
close in favor of #4384

PR #4420, opened 2023-11-24
PHRAS-2069 rescan file metadata command

aynsix:
### Adds
  - PHRAS-2069: rescan file metadata command
  - bin/console records:rescan-files-metadata

### DIFF
--- a/lib/Alchemy/Phrasea/Command/Record/RescanFilesMetadata.php
+++ b/lib/Alchemy/Phrasea/Command/Record/RescanFilesMetadata.php
@@ -18,6 +18,9 @@ class RescanFilesMetadata extends Command
     /** @var PhraseanetExtension */
     private $helpers;
 
+    /** @var \unicode */
+    private $unicode;
+
     public function __construct()
     {
         parent::__construct('records:rescan-files-metadata');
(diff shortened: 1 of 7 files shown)


lib/Alchemy/Phrasea/Command/Record/RescanFilesMetadata.php
```diff
@@ -0,0 +1,302 @@
...
+            $sql = "SELECT * FROM record WHERE " . $sql_where;
+
+            $stmt = $databox->get_connection()->prepare($sql);
+
+            if ($dry) {
+                $output->writeln("<info>" . $sql . "</info>");
+            }
+
+            $stmt->execute();
+            $rows = $stmt->fetchAll(\PDO::FETCH_ASSOC);
```
jygaulier, line 183: beware of big databoxes : better read row by row, and get only the needed columns

lib/Alchemy/Phrasea/Command/Record/RescanFilesMetadata.php
```diff
@@ -0,0 +1,302 @@
...
+                if ($results != NULL) {
+                    $metadatas = [];
+                    foreach ($results as $result) {
+                        $metadatas[] = [
+                            'action'         => $action,
+                            'meta_struct_id' => $metaStructId,
+                            'meta_id'        => null,
+                            'value'          => $result
+                        ];
+                    }
```
jygaulier, line 224: for multi-v, it is also possible to use value = [a,b,c] : use "results" directly in a single action.

lib/Alchemy/Phrasea/Command/Record/RescanFilesMetadata.php
```diff
@@ -0,0 +1,339 @@
...
+                    $action = 'add';
+                }
+            }
+
+            $output->writeln(sprintf("<comment>Working on database %s with Id : %d</comment>", $sbasName, $sbasId));
+
+
+            $sql_where = join(" AND ", $clauses);
+
+            $sql = "SELECT * FROM record WHERE " . $sql_where;
```
jygaulier, line 177: don't use * ;
**overwrite = null** optim: `... left join metadata... where ... and isnull(metadata)` so the select will not fetch records with value(s).

PR #4430, opened 2023-12-01
PHRAS-3957: editing: search in thesaurus, from the th zone

aynsix:
### Adds
  - PHRAS-3957: editing: search in thesaurus, from the th zone


Phraseanet-production-client/src/components/record/recordEditor/index.js
```diff
@@ -213,6 +215,11 @@ const recordEditorService = services => {
                     $editDateArea.css('width',210);
                 }
             })
+            .on('mouseup mousedown keyup keydown', '.editor-thesaurus-search', (event) => {
```
jygaulier, line 218: maybe only on keyup ? to be user-tested to see if refresh on click is expected or annoying...

PR #4443, opened 2023-12-14
PHRAS-3979: fix bin/maintenance cleaner 

aynsix:
### Fixes
  - PHRAS-3979: fix bin/maintenance cleaner

### DIFF
--- a/lib/Alchemy/Phrasea/Command/Maintenance/CleanLogDocsCommand.php
+++ b/lib/Alchemy/Phrasea/Command/Maintenance/CleanLogDocsCommand.php
@@ -94,7 +94,7 @@ class CleanLogDocsCommand extends Command
                 if ($dry) {
                     // for delete action, delete all event for the records
                     if (in_array('delete', $action)) {
-                        $sqlRecordIds =  'SELECT record_id FROM log_docs WHERE ' . $clauseWhere;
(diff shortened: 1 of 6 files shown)


lib/Alchemy/Phrasea/Command/Maintenance/CleanLogDocsCommand.php
```diff
@@ -128,16 +138,21 @@ public function doExecute(InputInterface $input, OutputInterface $output)
                         ->render($output);
 
                 } else {
-                    if (in_array('delete', $action) && !empty($rows)) {
-                        $recordsId = array_column($rows, 'record_id');
-                        $sqlDeleteAction = 'DELETE FROM log_docs WHERE record_id IN(' . implode(',', $recordsId) . ')';
+                    if (in_array('delete', $action)) {
+                        $sqlRecordIds =  'SELECT record_id FROM log_docs WHERE ' . $clauseWhere;
+                        $sqlDeleteAction = 'DELETE FROM log_docs WHERE record_id IN(' . $sqlRecordIds . ') LIMIT 1000';
```
jygaulier, line 143: if the first select returns 1 million lines, the IN(...) filter will explose mysql server.
we need 2 loops, each with limit 1000.
Apply this on every IN(SELECT())

aynsix: ok

PR #4462, opened 2024-01-24
PHRAS-4007: Auth provider - add keycloak provider - openid

aynsix:
### Adds
  - PHRAS-4007: Auth provider - add keycloak provider - openid


lib/conf.d/configuration.yml
```diff
@@ -228,26 +228,24 @@ authentication:
...
-                model-upfx: '_U_'
+                base-url: 'https://keycloak.phrasea.local'
+                realm-name: phrasea
+                icon-uri: null
+                birth-group: _firstlog
+                everyone-group: _everyone
+                metamodel: _metamodel
+                model-gpfx: _M_
+                model-upfx: _U_
+                debug: false
```
jygaulier, line 247: add `exclusive: false` as default
(add also in migration script)

aynsix: ok

PR #4475, opened 2024-02-15
PHRAS-4026 : js error on "change password"

aynsix:
### Fixes
  - PHRAS-4026 : js error on "change password"


jygaulier:
ok to me

PR #4497, opened 2024-04-17
PHRAS-4041:Prod - matomo - Field display - improve rendering of Matomomediametrics

aynsix:
### Fixes
  - PHRAS-4041: Prod - matomo - Field display - improve rendering of Matomomediametrics


jygaulier:
ok to me.
todo next (?) choose fields to display in conf ? - see with @nmaillat

PR #4531, opened 2024-07-09
PHRAS-4084 : workers - cancel a workers jobs if error is "record not found"

aynsix:
### Fixes
  - PHRAS-4084: workers - cancel a workers jobs if error is "record not found"

### DIFF
--- a/lib/Alchemy/Phrasea/WorkerManager/Command/WorkerRunServiceCommand.php
+++ b/lib/Alchemy/Phrasea/WorkerManager/Command/WorkerRunServiceCommand.php
@@ -39,14 +39,14 @@ class WorkerRunServiceCommand extends Command
             if ($body === false) {
                 $output->writeln(sprintf('<error>Unable to read payload file %s</error>', $input->getArgument('body')));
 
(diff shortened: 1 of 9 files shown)


lib/Alchemy/Phrasea/WorkerManager/Command/WorkerRunServiceCommand.php
```diff
@@ -39,14 +39,14 @@ protected function doExecute(InputInterface $input, OutputInterface $output)
...
-                return;
+                return 0;
             }
 
             $body = json_decode($body, true);
             if (json_last_error() !== JSON_ERROR_NONE) {
                 $output->writeln('<error>Invalid message body</error>');
 
-                return;
+                return 0;
```
jygaulier, line 49: why not returning a error code ? (1 ?) see: [https://hpc-discourse.usc.edu/t/exit-codes-and-their-meanings/414](https://hpc-discourse.usc.edu/t/exit-codes-and-their-meanings/414)

same thing todo for every failure point (catch's)

PR #4572, opened 2025-01-09
Delete www/.htaccess

gjacobjn:
### Removes
  - Remove .htaccess inside www


nmaillat:
close in favor of #4573

PR #4592, opened 2025-03-07
PHRAS-4138 empty field restricted thesaurus

aynsix:
### Adds
  - PHRAS-4138: empty field restricted thesaurus

### DIFF
--- a/Phraseanet-production-client/src/components/record/recordEditor/index.js
+++ b/Phraseanet-production-client/src/components/record/recordEditor/index.js
@@ -38,7 +38,6 @@ const recordEditorService = services => {
     let $searchThesaurusOperator;
     let $toolsTabs;
     let $idExplain;
-    let $buttonEmptyField;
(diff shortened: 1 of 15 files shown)


Phraseanet-production-client/src/components/record/recordEditor/index.js
```diff
@@ -640,8 +654,10 @@ const recordEditorService = services => {
 
                         if (field.input_disable) {
                             $editDateArea.prop('disabled', true);
+                            $buttonEmptyField.show();
```
jygaulier, line 657: The button can be useful even if the field is not "restricted" ; The only reason to hide the button is if the field is empty (do not do if too complicated).

Phraseanet-production-client/src/components/record/recordEditor/index.js
```diff
@@ -222,6 +224,18 @@ const recordEditorService = services => {
 
                 onUserInputComplete(event, $searchThesaurus.val(), currentField);
             })
+            .on('click', '.empty-field', event => {
+                event.preventDefault();
+                let fieldIndex = options.fieldCollection.getActiveFieldIndex();
+                let field = options.fieldCollection.getFieldByIndex(fieldIndex);
+
+                if (field.multi) {
+                    if (!confirm("Are you sure, this will remove all values if exist!")) {
```
jygaulier, line 233: translate ?

